PR #3, opened 2016-02-19
FIA field style updates

james-nesbitt:
This patch converts the views row handler to be based on the EntityRow handler, in order to get the rendered node into the FIA article body.  Also some clean up on the template variables, and removal of the fia node page from the view.

### DIFF
--- a/facebook_instant_articles.info.yml
+++ b/facebook_instant_articles.info.yml
@@ -1,9 +1,11 @@
-name: Facebook instant articles
+name: 'Facebook instant articles'
 type: module
-description: FB Instant articles RSS feed generator for integration into the FB IA program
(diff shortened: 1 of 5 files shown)


facebook_instant_articles.info.yml
```diff
@@ -1,9 +1,11 @@
-name: Facebook instant articles
+name: 'Facebook instant articles'
 type: module
-description: FB Instant articles RSS feed generator for integration into the FB IA program
+description: 'FB Instant articles RSS feed generator for integration into the FB IA program'
 core: 8.x
 package: Facebook
 features: true
 dependencies:
```
dbosen, line 7: What is the explicit numbering for?

PR #4, opened 2016-02-22
FIA: feed output reformat to RSS/XML

james-nesbitt:
Following suggestions in feedback from FB, I've reformatted the feeds view display, to use the Drupal core RSS style handler, and reformatted the rows.

The template reformatting is based on the following sample: https://developers.facebook.com/docs/instant-articles/publishing#sample-feed


james-nesbitt:
closing this.  I made a mistake in the branches used.

PR #6, opened 2016-02-22
FIA template fixes for author, link and language

james-nesbitt:
Some discovered template bugs after the refactor for the RSS/XML output.


dbosen:
You remove the  $header array  in src/Plugin/views/row/FiaFields.php, but the varuiable is used later on, maybe you should keep the initialisation.

james-nesbitt:
the variable should have been removed entirely.

The previous approach used a custom "header" template variable to pass FIA semantic header data to the template.  I switched to using "options" which directly connect to plugin options.  Using options allows much easier future integration with plugin options, which also have the options to use a plugin form.

I have updated the PR to remove "header" references/

PR #13, opened 2016-03-28
Issue #2694855: views filter for marked content entities

steffenr:
Issue on d.o. https://www.drupal.org/node/2694855


modules/fb_instant_articles_views/views/fb_instant_articles_views.views.inc
```diff
@@ -20,6 +21,16 @@ function fb_instant_articles_views_views_data_alter(&$data) {
       'handler' => 'views_handler_filter_valid_fia'
     ),
   );
+
+  $data['node']['fb_instant_articles_selected_views'] = array(
+    'title' => t('Selected Content for Facebook instant articles feed'),
+    'filter' => array(
+      'title' => t('Selected Content for Facebook instant articles filter'),
+      'help' => t('Filter for items that where selected for beeing shown in facebook instant articles feed.'),
```
m4olivei: Nitpick, spelling: "Filter for items that were selected for inclusion in a Facebook Instant Articles feed."

modules/fb_instant_articles_views/views/fb_instant_articles_views.views.inc
```diff
@@ -20,6 +21,16 @@ function fb_instant_articles_views_views_data_alter(&$data) {
       'handler' => 'views_handler_filter_valid_fia'
     ),
   );
+
+  $data['node']['fb_instant_articles_selected_views'] = array(
+    'title' => t('Selected Content for Facebook instant articles feed'),
+    'filter' => array(
+      'title' => t('Selected Content for Facebook instant articles filter'),
```
m4olivei: Nitpick: "Selected Content for Facebook Instant Articles _feed_"

m4olivei:
:+1:

modules/fb_instant_articles_views/views/handler/views_handler_filter_enabled_fia.inc
```diff
@@ -0,0 +1,72 @@
...
+      '#suffix' => '</div>',
+    );
+  }
+
+  /**
+   * Add this filter to the query.
+   */
+  function query() {
+    // We don't extend the query if the setting for all_enabled_default is set.
+    $default_enabled = variable_get('fb_instant_articles_all_enabled_default', TRUE);
```
m4olivei, line 58: Hum, this variable no longer exists after the refactor.  It's new name is `fb_instant_articles_rss_all_enabled_default` and it's a part of the `fb_instant_articles_rss` module.  The new thinking in the refactor was the `fb_instant_articles_rss` implements a dead simple RSS feed that internalizes everything it needs for the rudimentary filtering it does, and in the views module, site builders, if they needed it, would define whatever filtering they need.  For instance on the site I'm working on, I've added a boolean field to our blog nodes for "Publish to Facebook Instant Articles", then I filter the view based on that.

As such, I don't think that this filter is necessary anymore.  At least, it doesn't make sense without the `fb_instant_articles_rss` enabled, and even then, the presumption is that you either use `fb_instant_articles_rss` or `fb_instant_articles_views` and not both :confused:.

Does that make sense?

scottrigby, line 58: I second that idea. We discussed the rationale here https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/17#discussion_r57811329 and here https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/17#discussion_r57810546

scottrigby, line 58: In fact I was thinking we may want to keep those modules from being enabled at the same time, by adding a `hook_requirements()` check in both modules (you would need to disable one to enable the other, and vice versa).

steffenr:
While creating the PR i had not looked at the current state of your refactoring works. As m4olivei pointed out, i also think we won't need that views filter anymore. Creating a custom boolean field in the appropriate content types solves the filtering issue in a "drupal way" ;). @scottrigby I don't think that we should keep both modules enabled at the same time. If someone wants to keep it real simple, we would enable the rss submodule. Otherwise we would use the views submodule and and benefit from a lot more flexibility.

If anyone is fine with the suggestions, i would close the PR, cause it's no longer relevant.

scottrigby:
@steffenr I agree :+1: Closing

PR #15, opened 2016-03-29
rename files and change facebook_ to fb_ in content where necessary

dbosen:


scottrigby:
@dbosen I haven't been able to help with the 8.x version yet due to our immediate needs for this in D7, but the re-architecture plan we've all agreed on was made with both 8.x & 7.x in mind, so as soon as I get a chance I want to help on 8.x with the effort to bring the two major versions to feature parity. At some point though – after we accomplish that in general – I think we should make all feature changes to 8.x first, then port downstream to 7.x. Something to discuss later… 

Anyway that is just to say this patch looks good from the namespace change perspective, but I'm not able to fully test it just yet (and we should plan how to do the module separation in 8.x to match #17 ).

PR #16, opened 2016-03-29
Issue #2686983: Features integration

m4olivei:
This adds support to export the flag that turns on FB IA opt in per entity + bundle pair.

See https://www.drupal.org/node/2686983

### DIFF
--- a/modules/fb_instant_articles_display/fb_instant_articles_display.install
+++ b/modules/fb_instant_articles_display/fb_instant_articles_display.install
@@ -176,34 +176,3 @@ function fb_instant_articles_display_schema() {
 
   return $schema;
 }
-
-/**
- * Update fb_instant_articles_entity_types schema.
- */
-function fb_instant_articles_display_update_7000() {
-  // Add the id column to fb_instant_articles_display_entity_types
-  $field = array(
-    'description' => 'The unique id of this entity type mapping.',
-    'type' => 'varchar',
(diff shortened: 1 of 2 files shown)


m4olivei:
I'll rebase this once https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/17 gets in.

scottrigby:
@m4olivei reminder to rebase now that #17 has been merged :smile: 

I would remove the update hook though, since we're still in very active development before any release has been made. After we issue releases, this would be important :) Otherwise I think this is a good idea to add.

Another note is that while we allow entity type configuration, a lot of the current code only supports nodes. This is something we can take up in a later issue, but I want to mention it here for conversation sake since you're refactoring the storage of that setting in this PR.

m4olivei:
@scottrigby yep yep, I did the rebase.  Yeah, good call on the update hook, I was in the wrong mindset when I wrote that up, I'll remove it.

Agreed regarding entity_type config. I keep coming across stuff called 'entity_type' but not really supporting anything but nodes.  Definitely could be a follow up down the road.  It would probably need someone who actually has a use case outside of nodes to be motivated to write it.

m4olivei:
Update hook removed, should be good to go.

m4olivei:
@scottrigby does this still look good to you?  Would be nice to have this, it's one less thing I have to do to get my test environment setup.

PR #17, opened 2016-03-29
FB instant articles split modules

scottrigby:
OK, this PR does the initial module splitting work outlined in [Module Architecture](https://docs.google.com/document/d/1dv9cr-vgPwLVgjTXCFR7VpVe_D-C3LtrYBrRjtELG1Y/edit#heading=h.n379d0f16t4c) section of our collaborative Google doc. 

This covers only splitting the existing functionality (placeholder for base module, display module, and rss module) – it does not yet add the new functionality (api, reports, or even the base module classes). 

I would like to merge this in sooner than later, because ever new commit since I started working on this (for example https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/14) creates a rebase nightmare, since nearly everything has been moved into submodules.

One note: the `fb_instant_articles_page_id` variable is currently only being used by the base module to set the metatag. This is a funny case, because other than that the base module would have no need for a settings form at all. We will however need that for the API. For now, I didn't re-add that an accessible settings form because I think I'd like to have a convo with everyone about the best place to handle the metatag helper function, and I think we will want to address that separately as we flesh out the rest of the API part of this puzzle. For now I'd just like to get our agreed upon structure in place, and then refine from there.

Given the caveats above, any immediate thoughts or concerns about merging this?


m4olivei:
Gah, I was doing this at the same time as you were :P.  You beat me to the punch.  Oh well, at least I'm well versed and should be able to do a good review.  Will comb it over now.

simonengelhardt:
I will look through this later today as well.

One quick note: I was thinking we would do PRs against the `7.x/fb-ia-sdk` feature branch and then merge in to the release branch when we feel ready enough for a pre-release (just commented something to the same effect [here](https://www.facebook.com/groups/1030797546992409/permalink/1037612319644265/?comment_id=1040724852666345&reply_comment_id=1041512292587601&comment_tracking=%7B%22tn%22%3A%22R%22%7D)). How does that sound? If you're more into this PR going directly into `7.x-1.x` I guess we could just rebase/merge that downstream to `7.x/fb-ia-sdk` afterwards.

scottrigby:
@m4olivei ha #whoops. Yeah that would be great – but this shows we're on the same page, which really helps. If I had known though I would have waited then reviewed your PR :smile_cat: but this works too.

@simonengelhardt Well, I think we should get this very basic split merged into `7.x-1.x` so that the PRs by other people can start fitting into place. The work you're doing (and that I have been doing too) on the API stuff I think can just layer on top of this. Does that sound good to you? Yeah the downstream idea you mentioned is what I'm thinking – essentially we can make a PR from your `7.x/fb-ia-sdk` branch against `7.x-1.x`. And in the meantime, maybe I can even try to integrate the work I'm doing as a PR against your `7.x/fb-ia-sdk` branch, so that we have something we feel is ready before opening the aforementioned PR against `7.x-1.x`. Is this making sense?

fb_instant_articles.module
```diff
@@ -8,28 +8,14 @@
...
-  $items['admin/config/services/fb-instant-articles/settings'] = array(
-    'title' => 'Settings',
-    'type' => MENU_DEFAULT_LOCAL_TASK,
-    'weight' => -1,
+    'title' => 'Facebook Instant Articles',
+    'description' => 'Settings related to Facebook Instant Articles.',
+    'page callback' => 'system_admin_menu_block_page',
+    'access arguments' => array('access administration pages'),
+    'file' => 'system.admin.inc',
+    'file path' => drupal_get_path('module', 'system'),
```
m4olivei, line 18: Shouldn't this be:

``` php
// ...
'page callback' => 'drupal_get_form',
'page arguments' => array('fb_instant_articles_admin_settings_form'),
'access arguments' => array('administer fb_instant_articles'),
'file' => 'includes/admin.inc',
// ...
```

As it is there is nothing rendered at `/admin/config/services/fb-instant-articles`.

modules/fb_instant_articles_display/includes/view_mode_layout.inc
```diff
@@ -15,28 +16,28 @@ function fb_instant_articles_layout_form(&$form, &$form_state) {
...
-  $layout = fb_instant_articles_get_node_layout_settings($bundle_name);
+  $layout = fb_instant_articles_display_get_node_layout_settings($bundle_name);
   $form['#fbia_layout'] = $layout;
 
-  fb_instant_articles_field_ui_fields($entity_type, $bundle_name, $view_mode, $form, $form_state);
+  fb_instant_articles_display_field_ui_fields($entity_type, $bundle_name, $view_mode, $form, $form_state);
 
   // Attach js.
-  $form['#attached']['js'][] = drupal_get_path('module', 'fb_instant_articles') . '/js/admin.js';
+  $form['#attached']['js'][] = drupal_get_path('module', 'fb_instant_articles_display') . '/js/admin.js';
```
m4olivei, line 25: There is a JS error on the Manage Display page for the Facebook Instant Articles view mode.  Could be due to script ordering?  Digging in.

modules/fb_instant_articles_display/includes/view_mode_layout.inc
```diff
@@ -117,7 +118,7 @@ function fb_instant_articles_add_regions(&$form, &$form_state) {
   foreach (element_children($table) as $name) {
     $row = &$table[$name];
     $row['#js_settings'] = array('rowHandler' => 'facebookInstantArticles');
```
m4olivei: Should be renamed to `facebookInstantArticlesDisplay`

modules/fb_instant_articles_display/includes/view_mode_layout.inc
m4olivei, line 25: Nevermind, I see why, missed re-naming, see below.

modules/fb_instant_articles_rss/fb_instant_articles_rss.module
```diff
@@ -0,0 +1,92 @@
...
+  $permissions['administer fb_instant_articles_rss'] = array(
+    'title' => t('Administer Facebook Instant Articles RSS'),
+  );
+  return $permissions;
+}
+
+/**
+ * Implements hook_node_operations().
+ */
+function fb_instant_articles_rss_node_operations() {
```
m4olivei, line 46: This and the function below should go into `fb_instant_articles_display`, since the mass update would also be useful for the views integration.

modules/fb_instant_articles_rss/includes/rss.inc
```diff
@@ -46,3 +46,43 @@ function fb_instant_articles_rss_page() {
 
   echo $rss->asXML();
 }
+
```
m4olivei, line 49: The `module_load_include` statement in `fb_instant_articles_rss_page` needs to change to:

```
module_load_include('inc', 'fb_instant_articles_rss', 'includes/FacebookInstantArticlesRSSSimpleXMLElement');
```

Also a couple instances of `FacebookInstantArticlesSimpleXMLElement` were missed in renaming to FacebookInstantArticlesRSSSimpleXMLElement.

An instance of `fb_instant_articles_get_article_nodes()` was missed in renaming to `fb_instant_articles_rss_get_article_nodes`.

Also the variable names were missed in renaming.

modules/fb_instant_articles_display/includes/view_mode_layout.inc
scottrigby: @m4olivei Ah yes, definitely missed renaming. will change. Thanks!

fb_instant_articles.module
scottrigby, line 18: @m4olivei Well we could do this several ways:
1. Currently in this PR I'm making a menu section placeholder, like the system module does in `system_menu()`, so that other modules can add children menu items (example: `admin/config`). Without the children items (in our case we anticipate child uri paths from our own submodules), there will be no children settings pages. This is what I meant in the initial PR comment – apart from the metatag helper (which i think can be done in a different way somewhere else, but we need to talk about it outside the scope of this PR) I'm not certain yet that the base module needs any settings form at all on it's own. I just left that admin.inc file where it is till we decide how to handle that one variable.
2. Alternately we could create one admin form all on the same page, and then have submodules form alter that to add new sections as needed (adding groupings etc to clarify what each set of configs are for).

I'm not dead set on how we do it, but I did the initial PR this way because it's less tightly coupled (and less unnecessary alters) and in that way I personally find the nicer option of the two. But that's something we can all decide together. But does this help clarify?

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -0,0 +1,777 @@
...
+ * Submit callback for node form.
+ */
+function fb_instant_articles_display_node_form_submit($form, &$form_state) {
+  $fb_instant_enabled = fb_instant_articles_display_is_article_type('node', $form['type']['#value']);
+  if ($fb_instant_enabled) {
+    if (!isset($form_state['values']['fb_instant_enabled'])) {
+      $form_state['values']['fb_instant_enabled'] = 0;
+    }
+    $value = $form_state['values']['fb_instant_enabled'];
+    $default_enabled = variable_get('fb_instant_articles_display_all_enabled_default', TRUE);
```
m4olivei: This variable was renamed to `fb_instant_articles_rss_all_enabled_default`, so the logic here doesn't work and despite checking the box, the value isn't saved to the DB and the article doesn't show up in the default feed.  This variable probably _should_ be defined and managed in the `fb_instant_articles_display`, which means it needs an admin page.  The way I was going to do the admin page was that each submodule would form alter what they needed into a single form.  I'm down with either approach.

modules/fb_instant_articles_display/includes/view_mode_layout.inc
scottrigby: OK fixed this one.

m4olivei:
OK reviewed all but `fb_instant_articles_views`.  I'll check a little later, for now :baby: duty.

fb_instant_articles.module
m4olivei, line 18: Ah OK, cool beans.  See my later comment, maybe the form_alter makes sense so we don't have a bunch of tiny admin forms, and there is just one place for all settings.  My $0.02 anyway.

scottrigby:
@m4olivei thanks for the review so far, and catching some things I missed (i'll keep in-place to your inline comments). Yeah for the views submodule the only functional change in this pr is the dependency in the info file (to match the plan in the google doc), so you haven't missed anything really :smile:

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby: Hm yes I missed that one - nice catch. So… this throws a small wrench in my desire to change as little of this existing code as possible in this PR. Personally, I think all these settings really do have more to do with RSS submodule functionality than the Display-related functionality (because they determine which entities will show in the feed page, whereas the Views submodule can let site builders choose any method they wish for selecting which items get added to the Views-controlled feed: a custom checkbox field, date ranges, or other complex selection logic). This is why I added these settings to the RSS module's settings form – but yes I agree it doesn't make sense to have this simple selection functionality in the Display module but it's configs in the RSS module (since the dependency is the other way round). By this logic, perhaps these functions that relate to simple selection of feed items should all move to the RSS module. Does this make sense?

modules/fb_instant_articles_rss/includes/rss.inc
scottrigby, line 49: @m4olivei Thanks, fixed!

modules/fb_instant_articles_rss/fb_instant_articles_rss.module
scottrigby, line 46: @m4olivei I'm not sure about that. See my logic in https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/17#discussion_r57810546. My assumption is that the Views submodule allows site builders to determine any simple or complex selection criteria they wish for what should display in the Views-controlled feed (whereas the simple RSS submodule really only has one way to do this, hence Views as a more advanced option between these two submodules that do basically the same thing).

modules/fb_instant_articles_display/includes/view_mode_layout.inc
scottrigby, line 25: BTW I fixed the renaming issue in the now hidden (outdated) diff comment below.

modules/fb_instant_articles_display/fb_instant_articles_display.module
m4olivei: I follow what your saying.  I had it in my head that the filter handler in the views submodule made use of this variable (`fb_instant_articles_display_all_enabled_default`) and the simple checkbox setting on each node.  Turns out it doesn't, it just uses the content type setting to filter the view by content type (https://github.com/scottrigby/module-fb_instant_articles/blob/fb_instant_articles_split_modules/modules/fb_instant_articles_views/views/handler/views_handler_filter_valid_fia.inc#L61).  Plus, I agree with you in that site builders will choose arbitrary methods of filtering, that's the power of views.  Indeed that's what we have in mind for Bravo.  So, I'm on board with moving these functions into the RSS module.

modules/fb_instant_articles_rss/fb_instant_articles_rss.module
m4olivei, line 46: You're right.  As I noted there, I had it in my head that the views integration used this data in a custom filter, which it currently does not, so all good.

modules/fb_instant_articles_display/fb_instant_articles_display.module
vgardner: Right, also that function name has changed to fb_instant_articles_display_get_article_entity_types(), so we should either change that for now or remove that preconfigured filter altogether.

m4olivei: Agreed on the rename. Probably no harm in leaving the filter, I don't think it conflicts with the re-jigging going here as is.  We could definitely revisit it as a follow up in terms of if it's worth hanging onto.

modules/fb_instant_articles_rss/fb_instant_articles_rss.module
```diff
@@ -0,0 +1,92 @@
...
+        'callback arguments' => array('fia_enabled' => 1),
+      ),
+      'remove_fia' => array(
+        'label' => t('Unpublish selected content from Facebook Instant Articles RSS feed'),
+        'callback' => 'fb_instant_articles_rss_mass_update',
+        'callback arguments' => array('fia_enabled' => 0),
+      ),
+    );
+  }
+  return $operations;
```
vgardner, line 65: I get this notice if $fia_all_enabled is set to TRUE.
`Notice: Undefined variable: operations in fb_instant_articles_rss_node_operations() (line 64 of /sites/all/modules/module-fb_instant_articles/modules/fb_instant_articles_rss/fb_instant_articles_rss.module).`

vgardner:
Good work @scottrigby and @m4olivei. This looks good to me and all works apart from the current comments. I think we can probably take some time discussing which module is responsible for what. Since this is an initial restructure, we should probably move forward with it after comments are addressed and revisit certain parts individually afterwards.

fb_instant_articles.module
m4olivei, line 18: For the record though, I wouldn't want to hold this up on this point.

modules/fb_instant_articles_rss/includes/rss.inc
```diff
@@ -0,0 +1,88 @@
...
+  $rss->addAttribute('version', '2.0');
+  $channel = $rss->addChild('channel');
+
+  // Add Channel information;
+  $channel->addChild('title', variable_get('fb_instant_articles_channel_title', variable_get('site_name', 'Default site name')));
+  $channel->addChild('description', variable_get('fb_instant_articles_channel_description', variable_get('site_slogan', t('Default site slogan.'))));
+  global $base_url;
+  $channel->addChild('link', $base_url);
+  $channel->addChild('language', variable_get('fb_instant_articles_channel_language', 'en-gb'));
```
m4olivei: The three above variables missed the rename.

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -0,0 +1,777 @@
...
+  }
+}
+
+/**
+ * Implements hook_theme().
+ */
+function fb_instant_articles_display_theme() {
+  $theme_path = drupal_get_path('module', 'fb_instant_articles_display') . '/theme';
+  return array(
+    'node__fb_instant_articles_display' => array(
```
m4olivei: No need for rename, should be `node__fb_instant_article`.  This theme hook is designed to override the template suggestion for the nodes rendered as `fb_instant_article` view mode

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -0,0 +1,777 @@
...
+      'template' => 'node--fb-instant-articles-display',
+      'path' => $theme_path,
+      'render element' => 'elements',
+    ),
+    'field__fb_instant_articles_display' => array(
```
m4olivei: For the same reason as above, should be `field__fb_instant_article`.

m4olivei:
OK I'm done reviewing, turned on the views module and found a couple other things.  All in all, great work @scottrigby!

fb_instant_articles.module
scottrigby, line 18: @m4olivei I agree, and I think we can sort this out afterwards (easy to have a convo on pros & cons of each option, and there's no harm in changing – so long as the required variables can be set).

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby: @vgardner @m4olivei OK guys, so it sounds like we are in agreement on that – I will go ahead and move all the functions relating to simple entity selection for the rss feed to the rss module… and will make sure the naming in those functions match. Stay tuned for updates…

scottrigby:
OK guys regarding this [now outdated diff comment thread](https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/17#discussion_r57915199) – @m4olivei @simonengelhardt – the per-entity simple rss feed selection should be decoupled in these commits: 735fd8a and 052e8b5. I had to pick some of the form work apart into new functions. I pinged @vgardner since he initially wrote that part, but looks like he may be busy Today. Can you check to see how that's looking? @m4olivei I'll address your other inline comments in-place.

modules/fb_instant_articles_rss/fb_instant_articles_rss.module
scottrigby, line 65: @vgardner I was going to leave things like this alone in this PR (if a bug already existed), but since you mentioned it here I just fixed this in 0a37d75

modules/fb_instant_articles_rss/includes/rss.inc
scottrigby: @m4olivei thx, fixed in ca33574.

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby: @m4olivei Ha ok, that was hasty on my part - there has been a lot to sort through on this initial splitting. I'll fix these, and also the template file names from your comment below.

m4olivei:
@scottrigby those two updates look and work great!

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby: OK fixed in 19bd0f2.

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby: Fixed in 19bd0f2.

scottrigby:
I took the liberty of refactoring `fb_instant_articles_rss_form_node_form_alter()` by reusing an existing function to clean up the type check and double foreach loop in 0bbcc38 (otherwise I am going to try to leave the rest of this code alone and let other refactoring happen outside of this initial large decoupling PR). Any other issues keeping us from merging?

scottrigby:
One other thing I think we need to do ASAP is update the README. I'll try to whip out an initial version of that, but I'd like to suggest doing that as a separate PR, so that this one can get in once we've sorted out the initial functional issues related to decoupling into submodules.

m4olivei:
@scottrigby looks good to me.  There is a PHP notice on creating new content, but OK to handle in a follow up:

```
Notice: Undefined property: stdClass::$nid in fb_instant_articles_rss_form_node_form_alter() (line 181 of /var/www/docroot/sites/all/modules/contrib/fb_instant_articles/modules/fb_instant_articles_rss/fb_instant_articles_rss.module).
```

scottrigby:
@m4olivei yeah that would be an existing bug in the code, because new nodes won't have an `nid` until the node is saved (in the submit handler). So this just needs to be wrapped in a check. I agree it should be a follow up - there are probably a lot of things like this, I've touched as little of the code as i can (though it may not look like it on first glance since this PR moves a lot around, the code itself is mostly as it was).

scottrigby:
I've also fixed the existing hook documentation.

m4olivei:
Everything looks good, all comments addressed from @vgardner and I as well as some extra bonus cleanup.  I'll file follow up issues on d.o for the couple things noted along the way.  

I'm going to go ahead and push the button.  Nice work @scottrigby :100:.

m4olivei:
Filed two follow-ups per our discussion throughout this review:

https://www.drupal.org/node/2697295
https://www.drupal.org/node/2697299

simonengelhardt:
Really nice work, @scottrigby, @vgardner, and @m4olivei. Glad we have the overall structure in place, so we can iterate on functionality and feature coverage.

I will start merging this into the `7.x/fb-ia-sdk` feature branch, so we can continue work on the SDK and API integration there, within the new module architecture.

scottrigby:
@simonengelhardt I totally agree, this was a great team effort! I'll connect with you Tomorrow about collaborating on that branch for the API related work :smile:

PR #19, opened 2016-03-31
Issue #2697985: Update README based on module restructure

scottrigby:
D.o issue [#2697985](https://www.drupal.org/node/2697985).

Follow up to https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/17.

Guys, this README is definitely more complicated than the 8.x one, but then again the architecture is now more complex (in order to accommodate more use cases). I took an initial stab at this, trying to help both site builders and seasoned developers understand what the different options are for at a high level. 

My thinking is each submodule's configuration options and other details can be added into their own README files. I didn't add those yet, since we have yet to add some of those new modules into this branch. I introduced them all here to reflect what's in the [Module Architecture section of our google doc](https://docs.google.com/document/d/1dv9cr-vgPwLVgjTXCFR7VpVe_D-C3LtrYBrRjtELG1Y/edit?usp=sharing), so I don't think this PR is ready to merge yet. But I wanted to put this out to the group for early feedback. Thoughts?


scottrigby:
Note that the markdown is a little more readable when parsed. See https://github.com/scottrigby/module-fb_instant_articles/blob/2697985_update_7x_readme/README.md

vgardner:
Hey @scottrigby, the instructions are clear and I think this is a good starting guide for users of this set of modules.
Few points:
- Since we're using a Markdown file, we should update our hook_help to use that extension. Also, we can take advantage of the markdown filter - https://www.drupal.org/node/161085#hook_help
- Can we separate "Dependencies" into "Requirements" and "Recommended" as per https://www.drupal.org/node/2181737?
- I guess Toubleshooting and FAQ sections will evolve and probably belong to each submodule
- We should add the Maintainers section with the supporting organisations too (last section on https://www.drupal.org/node/2181737)

README.md
```diff
@@ -0,0 +1,120 @@
...
+The Drupal Facebook Instant Articles modules have different dependencies, see 
+"About Drupal integration" above to help choose which modules are right for you.
+
+- Base (fb_instant_articles.module)
+    - Depends on:
+        - [Facebook Instant Articles PHP SDK](https://github.com/facebook/facebook-instant-articles-sdk-php) 
+provides necessary support for ensuring Drupal content is added to facebook in 
+the required format for Instant Articles. 
+    - Optionally (recommended) integrates with:
+        - [Composer Manager](https://www.drupal.org/project/composer_manager) for
```
scottrigby: Make this more explicit, so users know they should enable composer manager unless they really REALLY know what they're doing (currently /Logger doesn't load and blows up).

scottrigby: @simonengelhardt This was updated in 4a8cb07.

scottrigby:
@vgardner re [your last comment](https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/19#issuecomment-205531937):
1. Good catch, the extension path for `hook_help()` is now fixed in 4c55ecb.
2. I'm open to that. But I did it this way so that we could have an overview of all the submodules on the main README (while details for each can be in their own README files as you said in your point 3 above). But maybe we should just list the dependencies and recommended modules on this README, and then add a line noting that each submodule has it's own README?
3. See above
4. How about instead we link to the [d.o comitters page](https://www.drupal.org/node/2676800/committers), since this is a wider community initiative? I did that for the authors line in [our main composer json file](https://github.com/BurdaMagazinOrg/module-fb_instant_articles/blob/7.x-1.x/composer.json#L7).

simonengelhardt:
@scottrigby Overall I think the README reads very well. Just a few responses to your latest comment:
1. how about implementing @vgardner's suggestion to pass through the markdown filter if installed? I just did a quick test of that (literally copying the code snippet from https://www.drupal.org/node/161085#hook_help) and it works really well (I'm happy to do a micro-PR afterwards, if you'd rather just merge this in)
2. I think that split sounds like a good idea.
3. 👍 
4. I think linking to the committers page is an elegant solution 👍

scottrigby:
@simonengelhardt OK started the rest of the points – in the main module and the views submodule. I'll split out the other dependency notes into a new README for each submodule when I get back to this.

vgardner:
Cool, thanks for the changes.

For the composer file it's fine, but authors/committers are not necessarily maintainers. A project can have various committers, but knowing who the maintainers are has also helped me in the past.

scottrigby:
@vgardner that is a good point. We could separate those into separate sections. Who then are the maintainers? Everyone in the [People section of our google doc](https://docs.google.com/document/d/1dv9cr-vgPwLVgjTXCFR7VpVe_D-C3LtrYBrRjtELG1Y/edit?pref=2&pli=1#heading=h.ykx9pnorw88) who have offered to help? This is a slightly different way of approaching maintainership than most Drupal modules. I am up for clarifying that, but I don't want to bikeshed this before the RC. Perhaps we can discuss that in a thread in our FB group, and update the README after we resolve that (likely after the RC)?

fb_instant_articles.module
```diff
@@ -77,9 +78,31 @@ function fb_instant_articles_help($path, $arg) {
...
+ * @param string $string
+ *   The unprocessed README file contents string.
+ *
+ * @return string|false
+ *   A string containing the markdown filtered text, or FALSE.
+ *
+ * @see fb_instant_articles_help()
+ */
+function fb_instant_articles_help_markdown($string) {
+  if (module_exists('markdown')) {
```
m4olivei, line 98: Heh, nice touch.

scottrigby, line 98: @m4olivei heh yeah, that was suggested by @vgardner. I was thinking we may want to add the admin help text into a md file of it's own (from [here](https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/33/files#diff-062ec483c58e7768ec4bac083c4b499fR61)), but we can follow up on that if we want to.

m4olivei:
Nice work all.  The base modules README makes sense of the many submodules in pretty simple language IMHO. :confetti_ball:

PR #21, opened 2016-04-01
Issue #2698727 by rromore: add support for different data-mode values…

robertromore:
… besides fullscreen

### DIFF
--- a/modules/fb_instant_articles_display/fb_instant_articles_display.module
+++ b/modules/fb_instant_articles_display/fb_instant_articles_display.module
@@ -471,11 +471,7 @@ function fb_instant_articles_display_field_formatter_settings_form($field, $inst
       $element['presentation'] = array(
         '#type' => 'radios',
         '#title' => t('Presentation style.'),
-        '#description' => t('See the "Presentation" option') . ' ' . l(t('here'), 'https://developers.facebook.com/docs/instant-articles/reference/image#options', array(
-          'attributes' => array(
-            'target' => '_blank',
-          ),
-        )) . ' ' . t('for more information.'),
+        '#description' => t("See the \"Presentation\" option under the <a href=\"@fbia-docs-image-presentation\">image section of Facebook's Instant Articles documentation</a> for more information.", array('@fbia-docs-image-presentation' => 'https://developers.facebook.com/docs/instant-articles/reference/image#options')),
         '#options' => array(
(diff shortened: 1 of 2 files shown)


modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -466,10 +468,21 @@ function fb_instant_articles_display_field_formatter_settings_form($field, $inst
...
         '#default_value' => $settings['comments'],
       );
-      $element['fullscreen'] = array(
-        '#type' => 'checkbox',
-        '#title' => t('Enable Fullscreen.'),
-        '#default_value' => $settings['fullscreen'],
+      $element['presentation'] = array(
+        '#type' => 'radios',
+        '#title' => t('Presentation style.'),
+        '#description' => t('See the "Presentation" option') . ' ' . l(t('here'), 'https://developers.facebook.com/docs/instant-articles/reference/image#options', array(
```
scottrigby: @robertromore This PR looks great. One thing, when registering stings with `t()` you should not concatenate them, as it makes translation difficult. In Drupal 7 it's considered best practice to use anchor tags within `t()` as needed, like this:

```
'#description' => t('See the "Presentation" option under the <a href="@fbia-docs-image-presentation">image section of Facebook's Instant Articles documentation</a> for more information.', array('@fbia-docs-image-presentation' => 'https://developers.facebook.com/docs/instant-articles/reference/image#options')),
```

If you ever need to add a `_blank` target attribute it's fine to add HTML in `t()` (directly inside the anchor tag in this case), though personally I wouldn't enforce that. There are external links all over the place inside Drupal's admin UI, so I trust the user to know how to navigate them as they choose. Instead I suggested a rewording above to clarify where the link will go, but that part is personal preference so you may want to adjust the message from my suggestion. Just food for thought on that point :smile:

robertromore: @scottrigby thanks for the suggestion! I've updated the PR to reflect it.

vgardner:
This looks good, though things will change a bit once #22 is merged as it adds the SDK implementation of FBIA Elements. Let's hold this one until that's in.

scottrigby:
@robertromore #22 is in now, so this can be rebased/updated to match the new SDK method.

m4olivei:
Closing as this is almost 5 yo now :grimacing:

PR #23, opened 2016-04-04
Issue #2699945: Rules module and Drupal core actions for API calls

anbarasan1985:
Note, This is for early discussion, and won’t be done until the API module is available. 
Working on adding Default rules.


anbarasan1985:
@scottrigby @m4olivei @vgardner Please review the module structure and implementation.

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.inc
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ * Action: Delete facebook instant article.
+ *
+ * @param object $entity
+ */
+function rules_action_fb_instant_articles_delete($entity) {
+  if (isset($entity->id)) {
+    $url = url(drupal_get_path_alias($entity->entity_type . "/" . $entity->entity_id), array('absolute' => TRUE));
```
m4olivei: Better to use `entity_uri()` rather than assume the makeup of the system path for entities.

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules_defaults.inc
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @file
+ * Provide default rules integration to manage facebook instant article using API.
+ */
+
+/**
+ * Implements hook_default_rules_configuration().
+ */
+function fb_instant_articles_api_rules_default_rules_configuration() {
+  //Handling add/update node article as published in FB Instant Articles library.
```
m4olivei: Nitpick: space needed after opening comment.  Same for all other inline comments.

m4olivei:
@anbarasan1985 seems fine to me.  Admittedly I'm not currently planning on using this sub-module, and it's been awhile since I've done rules work.

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.info
```diff
@@ -0,0 +1,5 @@
+name = Facebook Instant Articles API Rules
+description = Provide Drupal rules to manage facebook instant article using API.
+package = Facebook Instant Articles
+core = 7.x
+dependencies[] = fbia
```
vgardner: This dependency should be fb_instant_articles_api and it also depends on the Rules module.

anbarasan1985: Updated the dependency now and this module can be used without `rules` module as well since we have a `hook_action_info` so which can be used with default `action` and `trigger` module.

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.rules.inc
```diff
@@ -0,0 +1,78 @@
...
+function fb_instant_articles_api_rules_integration_access($type, $name) {
+  return user_access('administer site configuration');
+}
+
+/**
+ * Action: Add facebook instant article.
+ *
+ * @param object $entity
+ */
+function rules_action_fb_instant_articles_create($entity) {
```
scottrigby: Since create/update is the same call, maybe group these two into one "import" callback?

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.rules.inc
```diff
@@ -0,0 +1,78 @@
...
+    // @todo Flesh out these calls when the API module is ready.
+  }
+}
+
+/**
+ * Action: Delete facebook instant article.
+ *
+ * @param object $entity
+ */
+function rules_action_fb_instant_articles_delete($entity) {
```
scottrigby: let's rename this to "remove"

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.info
```diff
@@ -0,0 +1,6 @@
+name = Facebook Instant Articles API Rules
+description = Provide Drupal rules to manage facebook instant article using API.
+package = Facebook Instant Articles
+core = 7.x
+dependencies[] = fb_instant_articles_api
+dependencies[] = fb_instant_articles_display
```
scottrigby: newline :wink:

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
```diff
@@ -0,0 +1,85 @@
...
+ *
+ */
+function facebook_instant_articles_api_remove_action($node, $context = array()) {
+  fb_instant_articles_api_rules_remove_article($node);
+}
+
+/**
+ * Implementation of a Facebook instant article import.
+ *
+ *@param object $node
```
scottrigby: Tiny thing. Space

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
```diff
@@ -0,0 +1,85 @@
...
+    $layout_settings = fb_instant_articles_display_get_node_layout_settings($node->type);
+    $wrapper = InstantArticleNodeWrapper::create($node, $layout_settings);
+    $node->fb_instant_articles_display_wrapper = $wrapper;
+    $node_view = node_view($node, $view_mode = 'fb_instant_article');
+    $import_article = drupal_render($node_view);
+    try {
+      Drupal\FBInstantArticles\DrupalClient::importArticle($import_article, $node->status);
+    }
+    catch(Facebook\Exceptions\FacebookAuthorizationException $e) {
+      watchdog_exception('facebook_instant_article', $e);
```
scottrigby: Any logging via the API calls should be added to the API module. See [fb_instant_articles_display_init()](https://github.com/BurdaMagazinOrg/module-fb_instant_articles/blob/7.x-1.x/modules/fb_instant_articles_display/fb_instant_articles_display.module#L8) for the current technique. This is a good thought though, because:
1. We should open a new issue on d.o to add Drupal logging relevant to the API module.
2. We should open another issue on d.o to move the `LoggerAppenderDrupal` class file, and perhaps the `fb_instant_articles_display_enable_transformer_logging` variable to a setting across the board for all of the instant articles modules (and then each submodule can implement it's own hook to register the relevant SDK loggers appropriate to each).

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
```diff
@@ -0,0 +1,85 @@
...
+ *@param object $node
+ */
+function fb_instant_articles_api_rules_remove_article($node) {
+  if (isset($node->nid)) {
+    $url = url(drupal_get_path_alias("node/" . $node->nid), array('absolute' => TRUE));
+    Facebook\InstantArticles\Client::removeArticle($url);
+  }
+}
+
+// @todo
```
scottrigby: Let's move this to the first (`fb_instant_articles_api_rules_action_info()`) docblock.

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.rules.inc
```diff
@@ -0,0 +1,59 @@
...
+    'base' => 'fb_instant_articles_api_rules_action_remove',
+  );
+  return $actions;
+}
+
+/**
+ * Facebook Instant Articles API integration access callback.
+ */
+function fb_instant_articles_api_rules_integration_access($type, $name) {
+  return user_access('administer site configuration');
```
scottrigby: Let's add a new permission. Check out `fb_instant_articles_display_permission()`, `fb_instant_articles_api_permission()` etc for the convention we're using (we may as well keep it consistent). :smile:

scottrigby: Or we can keep it simple and not add an access callback (which just uses Rules existing perms).

scottrigby:
@anbarasan1985 other that those few comments this looks :+1: 

Another integration note: I'm currently reworking the Display module's transformer class, so I'd suggest we merge this as-is now (because it should work with the existing code in `7.x-1.x`), and then in that follow-up PR I'll make I will need to update this code at the same time as I update the Display module's code.

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
anbarasan1985: @scottrigby  Added to the d.o issue [#2702199](https://www.drupal.org/node/2702199) and [2702201](https://www.drupal.org/node/2702201)

anbarasan1985:
@scottrigby Noted, Review again and merge it if everything looks fine now.

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.info
scottrigby: This is done

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.rules_defaults.inc
```diff
@@ -0,0 +1,48 @@
...
+ */
+function fb_instant_articles_api_rules_default_rules_configuration() {
+  // Handling add/update node article as published in FB Instant Articles library.
+  $rule = rules_reaction_rule();
+  $rule->label = t('Add/update as published in FB Instant Articles library ' .
+                   'when a published Article node is saved');
+  $rule->active = FALSE;
+  $rule->event('node_presave')
+    ->condition('data_is', array('data:select' => 'node:status', 'value' => TRUE))
+    ->condition('data_is', array('data:select' => 'node:type', 'value' => 'article'))
```
simonengelhardt: It would be nice if this value was not hardcoded to `article`, but relied on the `fb_instant_articles_display_is_article_type` function (I hope that's possible). I just discussed this with @scottrigby so he can maybe add more context if needed.

scottrigby: Hey yep, instead of checking "Article" content type (which is entirely my fault, since it was my suggestion to do this :wink: ) we could use this function since they already configured it elsewhere. Of course, with Rules people could still add any custom conditions they want, and with Drupal core actions they could add any custom triggers they want… but it would be nice to have this option in the default (example) rules.

So it would be nice to make a new Drupal core trigger (for the core Action), and also a Rules module condition.

simonengelhardt: Another thought: Wouldn't it be a little more terse to use the `node_is_of_type` condition instead of `data_is`?

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.rules_defaults.inc
```diff
@@ -0,0 +1,48 @@
...
+ * Implements hook_default_rules_configuration().
+ */
+function fb_instant_articles_api_rules_default_rules_configuration() {
+  // Handling add/update node article as published in FB Instant Articles library.
+  $rule = rules_reaction_rule();
+  $rule->label = t('Add/update as published in FB Instant Articles library ' .
+                   'when a published Article node is saved');
+  $rule->active = FALSE;
+  $rule->event('node_presave')
+    ->condition('data_is', array('data:select' => 'node:status', 'value' => TRUE))
```
simonengelhardt: Line could be replaced with:

`->condition('node_is_published')`

(seems a little cleaner to me)

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.rules_defaults.inc
```diff
@@ -0,0 +1,48 @@
...
+
+  $configs['fb_instant_articles_rules_default_published'] = $rule;
+
+  // Handling add/update node article as unpublished in FB Instant Articles library.
+  $rule = rules_reaction_rule();
+  $rule->label = t('Add/update as unpublished in FB Instant Articles library ' .
+    'when a unpublished Article node is saved');
+  $rule->active = FALSE;
+  $rule->event('node_presave')
+    ->condition('data_is', array('data:select' => 'node:status', 'value' => FALSE))
```
simonengelhardt: Line could be replaced with:

`->condition(rules_condition('node_is_published')->negate())`

(seems a little cleaner to me)

scottrigby:
@anbarasan1985 @simonengelhardt I think this looks great. All comments have been addressed, so I'm going to merge this. If we find additional improvements or bugs, we can make follow up d.o issues / GitHub PRs. Nice work! :shipit:

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
```diff
@@ -0,0 +1,85 @@
...
+ * @param object $node
+ */
+function fb_instant_articles_api_rules_import_article($node) {
+  if (isset($node->nid)) {
+    $layout_settings = fb_instant_articles_display_get_node_layout_settings($node->type);
+    $wrapper = InstantArticleNodeWrapper::create($node, $layout_settings);
+    $node->fb_instant_articles_display_wrapper = $wrapper;
+    $node_view = node_view($node, $view_mode = 'fb_instant_article');
+    $import_article = drupal_render($node_view);
+    Drupal\FBInstantArticles\DrupalClient::importArticle($import_article, $node->status);
```
simonengelhardt, line 71: This will not work as the method can not be called statically. We need to create an instance of `DrupalClient` to get the configured credentials. I will fix this, test, and submit a new PR.

scottrigby, line 71: Thanks @simonengelhardt nice catch. Something like this?

```
\Drupal\FBInstantArticles\DrupalClient::create()->importArticle($import_article, $node->status);
```

simonengelhardt, line 71: @scottrigby yeah, something like that. But hang on a bit and let me fix, since I'm reworking a few things here (such as the fact that the D7 registry autoloader actually doesn't support namespaces).

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * Implementation of a Facebook instant article remove.
+ *
+ * @param object $node
+ */
+function fb_instant_articles_api_rules_remove_article($node) {
+  if (isset($node->nid)) {
+    $url = url(drupal_get_path_alias("node/" . $node->nid), array('absolute' => TRUE));
```
anbarasan1985, line 82: @simonengelhardt As you are making some changes, do you like to change this as below,
`$url = url('node/' . $node->nid, array('absolute' => TRUE));` . Since URL is acting as key between the site and FBIA, it will get changed every time we update the content, so I think we have to pass the real URL instead of aliased one.

scottrigby, line 82: @anbarasan1985 Can you make this a separate issue on d.o, which could then lead to a separate PR if we go that route? I think the URL alias change is something we can work on / discuss separately from what Simon is working on.

scottrigby, line 82: How about an issue title like:

> Use drupal internal path rather than alias

and description like:

> URL alias can change, but FB Instant Articles uses canonical URL as key for API interaction. Therefore we need to pass an unchanging canonical URL.

m4olivei, line 82: @anbarasan1985 if you're wanting the unaliased path, you'd need to do:

```
$url = url('node/' . $node->nid, array('absolute' => TRUE, 'alias' => TRUE));
```

As counter-intuitive as that option is, setting `alias` to TRUE tells `url()` that what your giving it is already an alias, so it won't bother looking it up.

anbarasan1985, line 82: @scottrigby @simonengelhardt Created an issue on d.o [#2702857](https://www.drupal.org/node/2702857)

PR #24, opened 2016-04-05
Issue #2700733: FB Instant Articles API module

scottrigby:
Here is an initial PR to add the API module. Currently it's almost only a single class (for API encapsulation), a config file, and permissions to manage that. The code is well documented, so it should be self explanatory. 

This PR is also required for https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/23


modules/fb_instant_articles_api/includes/admin.inc
```diff
@@ -0,0 +1,40 @@
...
+  $args = array(
+    '@development_mode' => 'https://developers.facebook.com/docs/instant-articles/api#create-fields',
+  );
+
+  // The Facebook app connection info could alternatively be stored via the fb
+  // module, but it's not yet ported to D8, and a bit cumbersome just for this.
+  $form['fb_instant_articles_api_app_id'] = array(
+    '#title' => t('Facebook App ID'),
+    '#type' => 'textfield',
+    '#default_value' => variable_get('fb_instant_articles_api_app_id'),
```
anbarasan1985, line 21: Can we change this to `facebook_app_id` so if the application already using the facebook functionality then they don't need to add it again. Same for secret - `facebook_app_secret`

scottrigby, line 21: @anbarasan1985  I suggest not, see my comment at line 16 for reasoning (also, the D7 fb module lets you add multiple apps… so we'd need to build more code to allow supporting that, which sadly I think would be more confusing than helpful. It's likely people will use a different fb app for instant articles than other fb integration).

scottrigby:
Hey gang, this is ready to re-review.

modules/fb_instant_articles_api/includes/DrupalClient.php
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * {@inheritdoc}
+   *
+   * Additionally try to catch an attempted import call.
+   *
+   * @todo Revisit this try/catch wrapper if the API gives an API option to see
+   *   if the page has passed review. This may in fact be the best way, but that
+   *   is still in discussion by the Facebook team.
+   */
+  public function importArticle($article, $takeLive = FALSE) {
```
anbarasan1985, line 50: Can we change this `$takeLive` to `$publsih`? so it will be self explanatory.

scottrigby, line 50: Hey @anbarasan1985 – So this is interesting, @simonengelhardt mentioned the FB Instant team has [already changed this in the API documentation](https://developers.facebook.com/docs/instant-articles/api#create), so it already accepts `published`, but it also still supports `take_live`. I suggest we stick to this for now, and update it when the SDK updates this param.

scottrigby, line 50: Note that this is why I added the SDK version in the base module composer file to this PR… I want to make sure that the API work we do matches the correct version of the SDK (for example, the `removeArticle()` method was added in `v0.6.0`). So, once the SDK param updates from `$take_live` to `$published`, we can make a new PR to update that here too, and change the SDK version in the composer file at the same time.

fb_instant_articles.module
```diff
@@ -18,6 +18,16 @@ function fb_instant_articles_menu() {
     'file path' => drupal_get_path('module', 'system'),
   );
 
+  $items['admin/config/services/fb-instant-articles/settings'] = array(
```
simonengelhardt, line 21: I can't seem to access this form (the URL just renders the same as the base `admin/config/services/fb-instant-articles`).

modules/fb_instant_articles_api/includes/DrupalClient.php
```diff
@@ -0,0 +1,63 @@
...
+   * {@inheritdoc}
+   *
+   * Additionally extend the parent method to simplify sitewide Drupal settings.
+   *
+   * @todo Wait for an answer from Facebook on the legal viability of using the
+   *   app_id|app_secret app static token method.
+   *
+   * See also @link https://developers.facebook.com/docs/facebook-login/access-tokens#apptokens Facebook documentation note about the app static token method. @endlink
+   */
+  public function create() {
```
simonengelhardt: This factory method should be `static`

modules/fb_instant_articles_api/includes/DrupalClient.php
```diff
@@ -0,0 +1,63 @@
...
+   *   is still in discussion by the Facebook team.
+   */
+  public function importArticle($article, $takeLive = FALSE) {
+    try {
+      parent::importArticle($article, $takeLive);
+    }
+    catch(FacebookAuthorizationException $e) {
+      // A Facebook authorization exception probably means the page has not yet
+      // passed review. In that case, try posting the article unpublished.
+      if (!$takeLive) {
```
simonengelhardt: Remove this `if`. `parent::importArticle` should be called with the second argument set to `FALSE` no matter what this method was called with (otherwise this `catch` doesn't actually achieve anything, but just repeats what happened above in `try`)

modules/fb_instant_articles_api/includes/admin.inc
```diff
@@ -0,0 +1,40 @@
...
+    '#type' => 'textfield',
+    '#default_value' => variable_get('fb_instant_articles_api_app_secret'),
+  );
+
+  $form['fb_instant_articles_api_development_mode'] = array(
+    '#title' => t('Facebook Instant Articles development mode'),
+    '#type' => 'checkbox',
+    '#default_value' => variable_get('fb_instant_articles_api_development_mode'),
+    '#description' => t('Articles published from a site in development mode can 
+      never be taken live. See <a href="@development_mode">development_mode</a>
```
simonengelhardt, line 35: A nit, but I would suggest adding `target="_blank"`

m4olivei:
Reads OK to me.  DrupalClient is handy for wrapping the configured creds.  I'm down with this.

modules/fb_instant_articles_api/includes/admin.inc
scottrigby, line 35: @simonengelhardt I suggest not doing that in Drupal admin settings links, since there are a number of them already without opening in new tabs automatically. It's not the end of the world if we do, but given that I trust the Drupal admins to know how to handle external links in the way that best suits them.

fb_instant_articles.module
scottrigby, line 21: @simonengelhardt  Oh, do you mean from the modules page? If so yep that's because I need to update the info file config path. Will do now, nice catch!

simonengelhardt, line 21: @scottrigby No, the Configure link on modules page leads to `admin/config/services/fb-instant-articles` which contains navigation to the config pages of the sub-modules and I think is what we want, right? But if I hit `admin/config/services/fb-instant-articles/settings` I don't get the settings form, but just that same navigation page.

scottrigby, line 21: @simonengelhardt interesting, I do see that :eyes: I just updated the config links though in ad92018, which seems better. I'm curious about the form issue you're seeing though. Maybe `drush cc menu`? (i assume you probably did that, but just in case).

modules/fb_instant_articles_api/includes/admin.inc
simonengelhardt, line 35: @scottrigby My thinking was simply that this is an external link and potentially annoying if I have put in some settings and then loose them because the link hijacks my tab (since I don't think there's any client-side change tracking on these settings pages). Is there a good reason not to open a new tab for an external link (again, this is really a nit, so I will leave it here and just let you decide 😉)?

modules/fb_instant_articles_api/includes/DrupalClient.php
scottrigby: Thanks @simonengelhardt fixed in 27dca96.

modules/fb_instant_articles_api/includes/DrupalClient.php
scottrigby, line 50: Note: I removed my earlier composer file commit d1c0a00 and rebased to keep Simon's composer file, and then built on top of that with dc08c57.

modules/fb_instant_articles_api/includes/DrupalClient.php
scottrigby: @simonengelhardt  Ha actually i meant that to be `if ($takeLive) {`, so that if the article was intended to post as unpublished, and the first time fails for reasons other than what we suspect, then we won't try to post unpublished twice. I'll fix that, but does this check make sense to you?

scottrigby: This should be fixed with cf14566 (at least in the way I describe. Also clarified the comment).

simonengelhardt: @scottrigby Yeah, that makes total sense. Thanks for fixing.

modules/fb_instant_articles_api/includes/admin.inc
scottrigby, line 35: @simonengelhardt yeah i can see that. Can we leave this question till we do a general admin UX review? But yeah, whatever we decide we should then make it consistent across all the admin forms.

fb_instant_articles.module
simonengelhardt, line 21: @scottrigby So sorry. I had cleared my menu cache, but it turns out I had messed up my sandbox and had a previous version of the module code lying around. Stupid.

I actually think the behavior on the Modules page made more sense before (having the Configure link lead to the interstitial `admin/config/services/fb-instant-articles`). Would you be up for reverting that first hunk of ad92018 and keep the second one?

scottrigby, line 21: Oh sure. That does sound reasonable… the link for the base settings (right now just the page ID) is to be found on that system blocks page anyway). Will do now

scottrigby, line 21: Done in 648c75c

PR #25, opened 2016-04-07
Issue #2691921 - Author field should be customizable

mkandelaars:
I originally posted this patch on the issue queue of "FB Instant Articles". Would be good to have some feedback as I'm proposing a slight change in how the module works and obtains the data to generate the FIA Feed. 

Originally posted at - https://www.drupal.org/node/2691921 
The value for author in the FIA Feed is currently the author (user) of the node, however this is not desirable for all sites. e.g. News websites may have a dedicated "contentcreator" user that puts up articles to a site, and lists the author of the article as a field in a node. In this instance, every article would appear to be by the same author "contentcreator" rather than its actual author.

My suggestion for a solution is to create an option for the output of Author to be either current behaviour (default), or to grab the value of Author from a views field. This would give complete control over what field to use for Author. This functionality could also be useful to set the value of Description too and possibly other fields too.

Cheers,
Mikey


m4olivei:
The 1.x branch is no longer supported. If this is still needed, please open up a new issue against the 3.x branch. Thanks!

PR #26, opened 2016-04-08
Issue #2702667: Cleanup the views row plugin

m4olivei:
Resolves https://www.drupal.org/node/2702667.

### DIFF
--- a/modules/fb_instant_articles_views/theme/fb_instant_articles_views.theme.inc
+++ b/modules/fb_instant_articles_views/theme/fb_instant_articles_views.theme.inc
@@ -37,4 +37,5 @@ function template_preprocess_views_view_row_fia(&$vars) {
   $vars['author'] = $item->author;
   $vars['created'] = $item->created;
   $vars['modified'] = $item->modified;
+  $vars['item_elements'] = empty($item->elements) ? '' : format_xml_elements($item->elements);
 }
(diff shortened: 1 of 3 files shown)


m4olivei:
@steffenr could you give this a look?  I believe you're using the views module right?

m4olivei:
Actually hang on a bit, I'm still working on refining this.

m4olivei:
OK, now it works. I also added in cores concept of have $item_elements, which themes/modules can alter in a preprocessor to add other tags to each item.  For example <description> isn't required, and there isn't a good way to pull that in a general sense, but with this, a module/theme could add that in.

scottrigby:
This looks good to me :+1: But I don't have time to test it Today. Can someone review/test and get back before we merge?

scottrigby:
Also @m4olivei let's hold off before actually merging. @vgardner and I were talking this morning about adding Drupal commit message style attribution inside of our GitHub merge commit messages, so people get credit on d.o for reviewing PRs (not just for the commits themselves). I would like to investigate automating the default GitHub commit message (maybe through a plugin we could add to this repo). We don't need to let that automation idea hold us up, because we can just emulate that manually, but I'd like to try to start doing that ASAP if it's easy enough :smile:

m4olivei:
@scottrigby by manually, do you mean doing the merge locally into `7.x-1.x` with a tailored commit message and then pushing to `7.x-1.x`?

scottrigby:
^ We sorted this out in FB chat :+1:

m4olivei:
@scottrigby and I chatted about this, but just to re-iterate, I've tested with this code and am currently using it on our site.  Working quite nicely.

PR #27, opened 2016-04-08
Issue #2702857 : Use drupal internal path rather than alias

anbarasan1985:
@scottrigby @simonengelhardt @m4olivei @vgardner Please review.

### DIFF
--- a/modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
+++ b/modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
@@ -79,7 +79,7 @@ function fb_instant_articles_api_rules_import_article($node) {
  */
 function fb_instant_articles_api_rules_remove_article($node) {
   if (isset($node->nid)) {
-    $url = url(drupal_get_path_alias("node/" . $node->nid), array('absolute' => TRUE));
+    $url = url("node/" . $node->nid, array('absolute' => TRUE, 'alias' => TRUE));
     Facebook\InstantArticles\Client::removeArticle($url);
   }
 }
--- a/modules/fb_instant_articles_rss/includes/rss.inc
+++ b/modules/fb_instant_articles_rss/includes/rss.inc
@@ -33,7 +33,7 @@ function fb_instant_articles_rss_page() {
   foreach ($nodes as $node) {
     $item = $channel->addChild('item');
     $item->addChild('title', $node->title);
-    $item->addChild('link', url(drupal_get_path_alias("node/" . $node->nid), array('absolute' => TRUE)));
+    $item->addChild('link', url("node/" . $node->nid, array('absolute' => TRUE)));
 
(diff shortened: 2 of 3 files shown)


scottrigby:
This looks :+1: to me. Anyone else care to review, @m4olivei I know you discussed it in the d.o issue.

simonengelhardt:
If I'm reading this correctly, this would only be right for sites that are configured to use the `node/id` form as their canonical URL (specified with `og:url` or `<link rel="canonical">`).

I am guessing that is not best practice (as an example, it would not work for [this article](http://www.telemundo.com/noticias/2016/04/07/arrecian-las-descalificaciones-entre-clinton-y-sanders)), so I would say we should remove `'alias' => TRUE` from these calls.

URLs will resolve to their canonical URL when shared on Facebook (the [URL Debugger](https://developers.facebook.com/tools/debug/) is a good tool for testing this resolve path through redirects and/or `og:url`/`<link rel="canonical">`).

anbarasan1985:
@simonengelhardt The idea here is every-time when we update the node the URL alias can change, but FB Instant Articles uses canonical URL as key for API interaction. Therefore we need to pass an unchanging canonical URL in order to handle content update/delete. Let us know if we understand something wrong.

simonengelhardt:
@anbarasan1985 I understand. The canonical URL is not just the key for API interactions, however, but also the key that determines whether a link is rendered as an Instant Article when shared on Facebook. So, the canonical URL for an Instant Article must be the canonical URL that the article has on the web (through the resolving chain I described above). You could say that the `node/id` form is canonical within Drupal, but to the web (Facebook, search engines, etc.) that is (for most sites, I would think) not the canonical URL. See this debugger lookup for an example: https://developers.facebook.com/tools/debug/og/object/?q=www.telemundo.com%2Fnode%2F1083040

Does that make sense?

anbarasan1985:
@simonengelhardt Yes, so if there is any path alias change on the content which is already created as an FBIA then do we need to delete the exiting one and re-add the new one?

simonengelhardt:
@anbarasan1985 A new one will have to be created, yes, which should just happen since our save hook hits the create+update endpoint. Also deleting the previous one would be ideal, but I would not say that has to be highest priority for an initial version of the module (since not doing it will just leave a "stale" IA in the library that can not be shared on Facebook (because the old URL either 404s or redirects to the new one, which the Facebook crawler follows) – the same thing happens if ingesting articles via RSS).

scottrigby:
@simonengelhardt @anbarasan1985 ah, I didnt' realize that FB would take care of this server-side in the case of a 404. If that's the case, then this makes perfect sense for the API. We should probably just close this one then?

anbarasan1985:
@simonengelhardt @scottrigby Ok sounds good, I'll close this PR now and Later we can work on deleting and re-creating the article when there is a PATH change

PR #29, opened 2016-04-09
Bring API client and module settings to MVP

simonengelhardt:
I have brought the API client and rules actions to a working state (most changes in 755602dc9b1a98714fd653fa3f02728a2abb70c6), where imports and deletes are working against my test Facebook Page.

Note that I took a simple approach to exposing the `instantArticle` field in `InstantArticleNodeWrapper`, since such an instance is what the API client operates on – probably particularly relevant to @scottrigby's ongoing reworking of this class.

I have also combined the two default API rules for published/unpublished. Since the node has that information contained within itself, I thought the most common use case would probably be to map this to the published state of the Instant Article. If a site requires a more involved setup, another rules setup could be implemented.

All settings, including FB authentication, were initially implemented by @chrisackermann in the Display submodule (just as a container). I have since moved and renamed settings to the appropriate submodules.


scottrigby:
@simonengelhardt There's a lot in this PR, so I'll comment as a list for now:
1. 675d580 looks great, and a good improvement to the API rules (we can probably separate that out and into a new d.o issue/PR).
2. The FB auth work (in ae7ad95, 055d86f, 7e7057b, 5ecb573), is that because the `app_id|app_secret` style static app token isn't going to work? I know we left this as a question last time we met, but I hadn't heard back (just not sure if the auth token is done here to match the wordpress plugin, or because the static token isn't going to work for FB Instant Articles at all).
3. The namespace changes in 4803d51 I think has now been resolved by https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/30.
4. The rest of the changes to the form are things I can help with too (rather than a custom JS file to do show/hiding etc, Drupal's Form API uses #states for most of that). There are also some other things I can help with to match Drupal's form standards. But beyond that, I'm wondering about some of the new form options (ads, metrics)… these don't do anything yet, do they (if so, maybe I just didn't realize how yet). 
5. Now that #30 is in, this PR needs a rebase against the current 7.x-1.x.

After that let me know what you think about the above, and I can help however makes sense :smile:

simonengelhardt:
@scottrigby Thanks for taking a look. I now there is a lot in here, but there were a lot of things in different places that didn't actually work in my end-to-end testing of an API Rules based setup, so this is the result of that. Let me go through your list (I've numbered, so we can reference if we need):
1. I realize now that this could have been contained in a separate PR, but since the rules module is still so new and given the time constraints I didn't want spend too much time with dividing things (though I did try to compose my commits somewhat logically). I can try to pick this commit out and submit a separate PR or we could, you know, merge it in with this one 😉
2. Yes. The static token (or any app token) does not allow for sufficient permissions to work with the Instant Articles API. We need a non-expiring page access token, that this flow provides. I tried to explain this in the commit message of 7e7057b ("_The Instant Articles API can not be accessed with an app token_"), but I admit that was a little terse.
3. Yeah. I'll rebase.
4. Sure. There are definitely things that can be refined (like the UI implementation details of show/hiding etc.). If you feel you have time to do that, we would appreciate that, but since this is functional, we could also refine/refactor afterwards. As for the ads and analytics options, you are right: the functionality that applies these to the `InstantArticle` instance still has not been implemented (I should have noted this). I am going to spend time on this tomorrow (will file d.o issues, so we can track properly) and add to this PR or a new one.
5. Will do.

Thanks again 👍

scottrigby:
@simonengelhardt awesome OK:
1. Agreed. I suggested that only because it could be merged immediately. Either way :)
2. OK dang. It was a good attempt… later it would be nice to understand why it can work for some apps but not others (maybe we should note this in the SDK somewhere, since that's supported in FB's main API docs). But given that, this optional UI is nice and easy to use :+1: 
3. :+1: 
4. I would like to help w that, or perhaps Anbu or anther Drupal dev in our working group can help with that on Monday before I make the tag (I'm thinking Monday EOD).
5. :+1:

simonengelhardt:
@scottrigby Now rebased on 7.x-1.x. 😅

Re 2. It can work for some apps but not for others because it depends on what the app does. Some actions on the Facebook Platform require permissions beyond what can be granted to an app access token. Publishing Instant Articles is one of those actions, since it requires the `pages_manage_instant_articles` permission. There is a good overview of the various types of access tokens in the Graph API docs [here](https://developers.facebook.com/docs/facebook-login/access-tokens), limititations of app access tokens [here](https://developers.facebook.com/docs/facebook-login/access-tokens#apptokens), and specifics of the Instant Articles API permissions [here](https://developers.facebook.com/docs/instant-articles/api#authentication).

Re 4. Great. I would suggest that we merge this PR in, since it is functional and ties together the configuration with actually being able to import and delete Instant Articles (that did not work before 97fff38b68b5ac73312b2a461d56063693859f6f). Then, we would have a shared starting point that is functional (unless your testing shows otherwise) and we can work together on improving the settings UI. How does that sound?

m4olivei:
@scottrigby @simonengelhardt Re 4, I had some time so I went ahead and re-jigged the settings page to be more organized like a typical Drupal admin.  Check out https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/33.  It's a PR against this branch, should be just a quick check, test and merge and then carry on with the more important bits of this PR :).

modules/fb_instant_articles_display/fb_instant_articles_display.info
```diff
@@ -2,6 +2,6 @@ name = Facebook Instant Articles Display
 description = Display module for Facebook Instant Articles.
 package = Facebook Instant Articles
 core = 7.x
-configure = admin/config/services/fb-instant-articles/display
 dependencies[] = ctools
 dependencies[] = fb_instant_articles
+scripts[] = js/admin.js
```
scottrigby, line 7: This is already added by `$form['#attached']['js']` in `fb_instant_articles_display_layout_form()`. You'd only add this line in the info file if you want this JS to load on every page.

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -5,6 +5,7 @@
  * Hook implementations for Facebook Instant Articles Display module.
  */
 
+
```
scottrigby, line 8: Let's remove this extra line (it was a remnant from 1814239).

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
```diff
@@ -62,7 +62,7 @@ class DrupalInstantArticleDisplay {
    *
    * @var InstantArticle
    */
-  private $instantArticle;
+  public $instantArticle;
```
scottrigby, line 65: Just curious, I see it was part of 97fff38 but why change $instantArticle from private to public?

modules/fb_instant_articles_api/src/DrupalClient.php
```diff
@@ -51,12 +54,12 @@ public function importArticle($article, $takeLive = FALSE) {
...
-      // passed review. In that case, try posting the article unpublished. Only
-      // try again if the article was intended to pe published, so we don't try
-      // to post unpublished twice.
-      if ($takeLive) {
+    catch(FacebookResponseException $e) {
+      // If this was an authorization exception and the error code indicates
+      // that the page has not yet passed review, try posting the article
+      // unpublished. Only try again if the article was intended to be
+      // published, so we don't try to post unpublished twice.
+      if ($e->getCode() === self::FB_INSTANT_ARTICLES_ERROR_CODE_PERMISSION && $e->getSubErrorCode() === self::FB_INSTANT_ARTICLES_ERROR_CODE_PAGE_NOT_APPROVED && $takeLive) {
```
scottrigby, line 62: :+1: sweet

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
simonengelhardt, line 65: Essentially because we need it in this line: https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/29/files#diff-26b2d4f4256779e153e253fbc312d48fR71

Makes sense?

scottrigby, line 65: Ah ok. Yes. This was off my radar because I was still trying to get the entire object to be an instance of the fb SDK InstantArticle class. Yep def need to keep it this way till (if) we rework that in #35 :+1:

includes/admin.inc
```diff
@@ -5,29 +5,222 @@
...
+    '#collapsible' => TRUE,
+    '#description' => t('Choose your preferred method for displaying ads in your Instant Articles and input the code in the boxes below. Learn more about your options for <a href="@ads_url">advertising in Instant Articles</a>', array('@ads_url' => 'https://developers.facebook.com/docs/instant-articles/ads-analytics')),
+  );
+
+  $form['ads']['fb_instant_articles_ad_type'] = array(
+    '#type' => 'select',
+    '#title' => t('Ad Type'),
+    '#default_value' => variable_get('fb_instant_articles_ad_type', FB_INSTANT_ARTICLES_AD_TYPE_NONE),
+    '#options' => array(
+      FB_INSTANT_ARTICLES_AD_TYPE_NONE => FB_INSTANT_ARTICLES_AD_TYPE_NONE,
```
scottrigby, line 79: If you want option keys and values to match, in D7 you can use `drupal_map_assoc()` on the array, or to prep for D8 you can use `array_combine($my_array, $my_array)` (see change [D8 record](https://www.drupal.org/node/2207453)).

scottrigby:
Given our RC1 goal for Tomorrow, I'm merging this because it works. We can make follow up issues on d.o to make the settings form more Drupal-standard, potentially change the `exit` calls to `drupal_exit()` (if that's needed, or revisit), and anything else that we notice afterwards. This is really great, and glad it's all working :+1:

modules/fb_instant_articles_display/fb_instant_articles_display.info
simonengelhardt, line 7: Yeah, this was a merge blunder (I had removed it here, but it came back during a rebase conflict resolution). I'll remove it.

includes/admin.inc
simonengelhardt, line 79: Thanks for that input. If you see my commit message [over here](https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/36/commits/b52ca3350f8076fe0dab4d49bad3f9ca09902803), I actually made a note that we should follow up an change this behavior a bit (so we don't store the constant values in db, but the constant names (or something more static)). I'll file a d.o issue on this.

PR #30, opened 2016-04-09
Issue #2703365: Resolve D7 class namespace

scottrigby:
This PR addresses what's described in d.o issue [Resolve D7 class namespace](https://www.drupal.org/node/2703365).

I'd like to get this in, so that I can build on top of it for the other class-related issues:
- [Refactor Transformer/InstantArticle classes to be reusable and extensible](https://www.drupal.org/node/2703063)
- [Move the LoggerAppenderDrupal class file](https://www.drupal.org/node/2702201)


scottrigby:
Note because it adds a new module dependency, you will have to install fresh, or else download and enable the [X Autoload module](https://www.drupal.org/project/xautoload) (or just `drush dl xautoload && drush en xautoload -y`).

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
```diff
@@ -2,9 +2,11 @@
 
 /**
  * @file
- * Defines the InstantArticleNodeWrapper class.
+ * Contains \Drupal\fb_instant_articles_display\DrupalInstantArticleDisplay.
  */
 
+namespace Drupal\fb_instant_articles_display;
```
m4olivei, line 8: With this class now being namespaced, need to use fully qualified name for `DateTime` class throughout, ie `\DateTime` (lines 97 & 103)

fb_instant_articles.info
```diff
@@ -4,3 +4,4 @@ package = Facebook Instant Articles
 core = 7.x
 php = 5.4
 configure = admin/config/services/fb-instant-articles
+dependencies[] = xautoload (>=7.x-5.x)
```
m4olivei: As discussed, turns out your can get what we need out of composer_manager module, so we won't need to depend on xautoload.

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
scottrigby, line 8: @m4olivei  Nice catch. Fixed in 1e288a4 (I did them all, not just within the namespaced classes).

fb_instant_articles.info
scottrigby: Agreed. Fixed in 3688e9a.

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -13,15 +13,15 @@
 */
 function fb_instant_articles_display_init() {
  // Configure Instant Articles Transformer logger to append to Drupal log if enabled
- $enable_transformer_logging = variable_get('fb_instant_articles_display_enable_transformer_logging', FALSE);
+ $enable_transformer_logging = variable_get('fb_instant_articles_display_enable_transformer_logging');
```
simonengelhardt, line 16: I think I would like the default to remain off, since the transformer logger is quite chatty on the DEBUG level (I think we may need to change that a little bit). Unless there's a Drupal convention to have modules enable as much logging as possible by default and let people dial it down if they want.

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
```diff
@@ -28,8 +30,11 @@
 /**
  * Facebook Instant Article node wrapper class.  Builds up an InstantArticle
  * object using field formatters.
+ *
+ * Class DrupalInstantArticleDisplay
+ * @package Drupal\fb_instant_articles_display
  */
-class InstantArticleNodeWrapper {
+class DrupalInstantArticleDisplay {
```
simonengelhardt, line 37: Now that this class is namespaced under `Drupal` it seems redundant to have that also in the name. I would suggest maybe naming something that is a little more descriptive. To me, this feels like an _Adapter_ more than a _Wrapper_ or a _Display_. Essentially it adapts/transforms a Drupal node to an InstantArticle, right? So we could consider calling it something like `Drupal\fb_instant_articles_display\NodeAdapter\`

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby, line 16: @simonengelhardt Oh I removed that since the default 2nd param for `variable_get()` is `NULL`, so it still works the same.

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
scottrigby, line 37: @simonengelhardt Ha yeah, I thought a similar thing. I agree it would be better to change that, and I actually meant to change this to `DisplayInstantArticle`. The reason is I'm also working on refactoring that with a new class in the base module that extends `Facebook\InstantArticles\Elements\InstantArticle`, and then this Display class will then extend the one in the base module.

modules/fb_instant_articles_display/fb_instant_articles_display.module
simonengelhardt, line 16: @scottrigby d'oh, of course 😄

m4olivei:
Looks good :100:.

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
simonengelhardt, line 37: @scottrigby yeah – this is not really important btw – my thinking was that in the fully qualified name `Drupal\fb_instant_articles\DrupalInstantArticleDisplay` the class name just repeats the words that are already in the namespace, so it doesn't really tell you what the class does. `Drupal\fb_instant_articles\DisplayInstantArticle` has the same problem. That's why I thought `NodeAdapter` or something similar could make sense.

But.. We can completely just get back to this later on. No need to hold off on this and I see the PR was just merged in 👍

m4olivei, line 37: Derp :/  sorry about that @simonengelhardt.  Didn't see your comments come in on the PR until after I merged it.

modules/fb_instant_articles_display/fb_instant_articles_display.module
m4olivei, line 16: Good point regarding the chatti-ness of the transformer.  It also logs lines of the same message with separate calls.  From what I've read, I'm not sure thats the intended use.  At least, it comes out kind of awkward in the Drupal DBLog module the way we have it setup, eg.

http://take.ms/y59qK

Since the Drupal watchdog displays the most recent log message first, you have to read the set of message from a single set, if you will, backwards.

We should probably do something about this, although, total post-F8 task.  Wondering if this should get fixed upstream in the SDK?

I'll volunteer to write up the issue in the right spot.

simonengelhardt, line 16: @m4olivei you're spot on: those are exactly the changes we'll have to make in the transformer logger: using a renderer (I think those can be customized in `log4php`) to group those lines in a single messages that make sense whether read chronologically or reverse. Feel free to file the issue in the SDK repo and we'll get right to it post the F8 launch.

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
simonengelhardt, line 37: No worries @m4olivei – the naming suggestion we can just save for later.

PR #31, opened 2016-04-09
Issue #2702201:  Move logging to base module

scottrigby:
This PR depends on #30


fb_instant_articles.module
```diff
@@ -83,3 +83,28 @@ function fb_instant_articles_html_head_alter(&$head_elements) {
...
 }
+
+/**
+ * Implements hook_init().
+ *
+ * @todo Update to use a custom DrupalFBInstantArticlesLogger once
+ *   @link https://github.com/facebook/facebook-instant-articles-sdk-php/issues/14 this SDK issue @endlink is in resolved.
+ */
+function fb_instant_articles_init() {
+  $enable_transformer_logging = variable_get('fb_instant_articles_enable_transformer_logging');
```
m4olivei: @scottrigby the variable was renamed to `fb_instant_articles_enable_logging` in the admin, so setting it in the admin doesn't work to turn it on.  The logger is currently setup for the `facebook-instantarticles-transformer` appender, so calling it `fb_instant_articles_enable_transformer_logging` makes sense.  However, I think your intent was to make this a "all things Facebook Instant Articles logging" flag, so this probably just needs to change to `fb_instant_articles_enable_logging` and then we're good to go.

scottrigby: @m4olivei I missed this one, thanks. Yes, the idea is to enable **all** logging. When/if we find more appenders in the Instant Articles SDK we can register them all here. That can be a follow-up as needed though. For now, fixed the variable in ec6e6cd.

m4olivei:
Looks good :+1:

PR #32, opened 2016-04-11
Update SDK dependency

simonengelhardt:
v1 of the SDK has been released. Let's depend on that as a minimum.


scottrigby:
:shipit:

PR #33, opened 2016-04-11
Re-jig the admin for the base module.

m4olivei:
- Use hook_help for the giant precursor text.  
- Use `#description` and simple fieldsets. 
- Use #states API and remove custom Javascript.
- Re-order fields in order of importance.
- Wrap all string in t().


simonengelhardt:
Nice work! Moving the setup blurb to `hook_help` was exactly what I was going to do, so thanks for beating me to it ;)

I've added a couple of very minor tweaks (hope you don't mind) and will merge in 🚢

PR #34, opened 2016-04-11
Issue #2704315: Analytics support

simonengelhardt:
Will include analytics tracking code if configured in settings (depends on settings implemented in #29)


scottrigby:
:shipit:

PR #36, opened 2016-04-11
Add support for ads

simonengelhardt:
Will include an automatically placed ad code of the specified type if configured in settings (depends on settings implemented in #29). Approach is very similar to #34.


src/AdTypes.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\fb_instant_articles_display\DrupalInstantArticleDisplay.
+ */
+
+namespace Drupal\fb_instant_articles;
+
+abstract class AdTypes {
```
scottrigby: @simonengelhardt is this so you can get constants both within the procedural form functions and from within the `addAdsFromSettings()` method? If so I can see this. Are these ad types likely to change?

includes/admin.inc
```diff
@@ -108,21 +105,21 @@ function fb_instant_articles_module_config_ads($form) {
...
 
-  $form['ads']['fb_instant_articles_ads_dimension'] = array(
+  $form['ads']['fb_instant_articles_ads_dimensions'] = array(
     '#type' => 'select',
     '#title' => t('Ad Dimensions'),
     '#options' => array(
       0 => t('Large (300 x 250)'),
     ),
-    '#default_value' => variable_get('fb_instant_articles_ads_dimension'),
+    '#default_value' => variable_get('fb_instant_articles_ads_dimensions'),
```
m4olivei: Need to rename the `variable_del` statement in the `.install` file as well.

simonengelhardt: Good catch. Will do.

m4olivei: Also, should be consistent with the default value here, going by the changes further down, should be:

```
variable_get('fb_instant_articles_ads_dimensions', AdTypes::FB_INSTANT_ARTICLES_AD_TYPE_NONE)
```

m4olivei: Actually, hang on, does `AdTypes::FB_INSTANT_ARTICLES_AD_TYPE_NONE` make sense with this?  Will comment on lines where I saw this.

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
```diff
@@ -550,4 +553,71 @@ private function fieldFormatSocialElement($items, $body) {
...
+
+  /**
+   * Add ads if configured in settings
+   */
+  public function addAdsFromSettings() {
+    $ad_type = variable_get('fb_instant_articles_ad_type', AdTypes::FB_INSTANT_ARTICLES_AD_TYPE_NONE);
+    if ($ad_type === AdTypes::FB_INSTANT_ARTICLES_AD_TYPE_NONE) {
+      return;
+    }
+    $dimensions = variable_get('fb_instant_articles_ads_dimension', AdTypes::FB_INSTANT_ARTICLES_AD_TYPE_NONE);
```
m4olivei: Does `AdTypes::FB_INSTANT_ARTICLES_AD_TYPE_NONE` make sense as a default here?

src/AdTypes.php
simonengelhardt: Yes, that's exactly why. I wouldn't say they are very likely to change.

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
```diff
@@ -550,4 +553,71 @@ private function fieldFormatSocialElement($items, $body) {
...
+    $width = 300;
+    $height = 250;
+    switch ($dimensions) {
+      case 0:
+        $width = 300;
+        $height = 250;
```
m4olivei: This is kinda odd, and not so future friendly.  What if the options for `$dimensions` followed a convention, like 300x250, which could be broken up to set the `$width` and `$height`.  Easier to define new options, and for a future where maybe we expose an alter hook for the oddball size needs?

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
simonengelhardt: I would say so, but you are absolutely right in [your other comment](https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/36#discussion_r59301324) that this default should be applied there as well.

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
```diff
@@ -550,4 +553,71 @@ private function fieldFormatSocialElement($items, $body) {
...
+        $an_placement_id = variable_get('fb_instant_articles_ads_an_placement_id');
+        if ($an_placement_id) {
+          $ad->withSource(
+            'https://www.facebook.com/adnw_request?' .
+            drupal_http_build_query(
+              array(
+                'placement' => $an_placement_id,
+                'adtype' => 'banner' . $width . 'x' . $height,
+              )
+            )
```
m4olivei: More Drupal-y:

``` php
url('https://www.facebook.com/adnw_request', array(
  'query' => array(
    'placement' => $ad_placement_id, 
    'adtype' => 'banner' . $width . 'x' . $height,
  ),
));
```

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
simonengelhardt: Yeah, we can do that instead 👍

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
```diff
@@ -550,4 +553,71 @@ private function fieldFormatSocialElement($items, $body) {
       $body->addChild($social);
     }
   }
+
+  /**
+   * Add ads if configured in settings
+   */
+  public function addAdsFromSettings() {
```
scottrigby, line 581: @simonengelhardt would this method make more sense in the base class? This isn't really specific to a Drupal fieldable entity (which to me is the scope that the Display submodule adds).

scottrigby, line 581: Actually never mind, let's save this as a potential follow-up issue, for when we resolve the inheritance vs composition class structure question.

m4olivei:
I'm getting a couple errors when rendering my feed.  Digging for more information:

```
Warning: DOMDocument::importNode(): Couldn't fetch DOMDocumentFragment in Facebook\InstantArticles\Elements\Element->dangerouslyAppendUnescapedHTML() (line 47 of /var/www/docroot/sites/all/vendor/facebook/facebook-instant-articles-sdk-php/src/Facebook/InstantArticles/Elements/Element.php).
Recoverable fatal error: Argument 1 passed to DOMNode::appendChild() must be an instance of DOMNode, null given, called in /var/www/docroot/sites/all/vendor/facebook/facebook-instant-articles-sdk-php/src/Facebook/InstantArticles/Elements/Ad.php on line 215 and defined in Facebook\InstantArticles\Elements\Element->dangerouslyAppendUnescapedHTML() (line 48 of /var/www/docroot/sites/all/vendor/facebook/facebook-instant-articles-sdk-php/src/Facebook/InstantArticles/Elements/Element.php).
```

src/AdTypes.php
scottrigby: So I'm asking because the use statement and class calls in the form seem like an awful lot for an array of four options. Maybe we just make a function in the module that defines the array? (I know we can define an array in a constant pre-5.6 using serialize, but that doesn't help you access it in the class and settings form whereas a simple function would). 

Something like this:

```
function fb_instant_articles_get_ad_types() {
  return array(
    'none' => 'None',
    'fban' => 'Facebook Audience Network',
    'source_url => 'Source URL',
    'embed_code' => 'Embed Code',
  );
}
```

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
simonengelhardt, line 581: @scottrigby Agreed. It would make sense to put this in the base class you have been working on in #35 (same thing goes for the analytics added in #34), so we should revisit there.

src/AdTypes.php
scottrigby: It would make the code a lot more readable as well.

m4olivei:
It seems to fine right after a drupal cache clear, but with a primed cache, I get the error.

m4olivei:
Hurm, so it seems my issue may be unique b/c we have `entitycache` module in play on our site.

m4olivei:
Damn, DOMDocumentFragment and DOMNode classes don't support being serialized, so in my case, what happens is, the $node get's loaded up, the ads from the settings get added to the display object including the reference to a DOMDocumentFragment for when you have a ad using embed code.  `entitycache` module serializes and caches the loaded $node and when it fetches from cache on the subsequent request, the DOMDocumentFragment is empty, and it throws errors.

Not blocking, b/c specific to `entitycache`, just noting.  Can be dealt with in a follow up issue.

scottrigby:
@m4olivei excellent sleuthing!

src/AdTypes.php
simonengelhardt: @scottrigby I was actually just trying to implement a simple enum and took a cue from the comments on the [`SplEnum` docs](http://php.net/manual/en/class.splenum.php). I liked having something that was statically replaced at compile time, so it was not possible to have an undiscovered typo in there.

Also, if I shortened the const names (i.e. `AdTypes::NONE`) I think readability would actually be pretty good (even worth the `use`).

That being said, something dynamic like your approach could mitigate alleviate the issues I mentioned in the message for b52ca3350f8076fe0dab4d49bad3f9ca09902803

I'll try to implement something like that instead.

modules/fb_instant_articles_display/src/DrupalInstantArticleDisplay.php
simonengelhardt: Actually.. It does not make sense in that line at all (my eyes were in the wrong place) 👓 I will fix that.

m4olivei:
Looks good, tests well, all comments addressed.  Great work @simonengelhardt!

m4olivei:
Also, filed a follow up for the `entitycache` module issue I found: https://www.drupal.org/node/2704651.

PR #39, opened 2016-04-12
Issue #2704695: Missing id column in INSERT statement for fb_instant_articles_display_entity_types table

m4olivei:
Otherwise the features export doesn't have a key value to use in the export.


scottrigby:
Ha whoops :+1: Well, this can be part of rc2 :wink:

PR #44, opened 2016-04-13
Issue #2705919: Rules module not giving complete node object

anbarasan1985:
We checked on the demo environment and this solve the issue that we faced.


scottrigby:
Yep, tested on the demo site. :shipit:

PR #45, opened 2016-04-14
Issue #2706581: PHP notice when mapped field is empty for some nodes in the feed

m4olivei:
Resolves https://www.drupal.org/node/2706581.


scottrigby:
Code review :+1:

PR #46, opened 2016-04-15
Issue #2706763: Allow the FBIA Image field formatter to have an option for original image

m4olivei:
Resolves https://www.drupal.org/node/2706763.


scottrigby:
:+1: :shipit:

PR #48, opened 2016-04-19
Issue #2709297: Missing </html> tag on wrapping markup for Transformer field formatter

m4olivei:
Resolves https://www.drupal.org/node/2709297.


scottrigby:
I tested and this fix doesn't break anything. Standard Drupal handles this anyway, but this is nicer to explicitly close the tag 👍

PR #50, opened 2016-04-20
Issue #2709539: Warning illegal string offset

ghost:
Fix for the Ilegal String Offset issue reported at https://www.drupal.org/node/2709539


scottrigby:
Nice 🐛  catch

PR #51, opened 2016-04-21
Issue #2711079: Image failed to import to Fb instant article without doing node load

anbarasan1985:
Ready to review.

### DIFF
--- a/modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.rules.inc
+++ b/modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.rules.inc
@@ -37,9 +37,6 @@ function fb_instant_articles_api_rules_rules_action_info() {
  * @param object $node
  */
 function fb_instant_articles_api_rules_action_import($node = NULL) {
-  // Loading this again as we are missing some field values and resetting cache
-  // to get the updated node object.
-  $node = node_load($node->nid, NULL, TRUE);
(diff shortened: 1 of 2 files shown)


scottrigby:
That's right, we initially saw this even on a default D7 Drupal install when using the API Rules module (that's why one of it's functions adds an extra node_load(). I like the approach in your patch better, because it targets the missing fully loaded file. But let's get some other people's 👀  on this before we commit it, since this affects not only the API calls but feeds as well.

scottrigby:
@m4olivei @vgardner Thoughts on this? My thinking ATM is in the worst case it's harmless (if the file `uri` exists, it's not re-loaded), but in the best case this solves an issue we saw under a few different conditions.

m4olivei:
Just curious, why is it missing `uri`?  Where is `uri` typically set?

anbarasan1985:
we don't know why it is missing the `uri` but it is setting the value if you do `node_load` again, so it is going to set using `file_load` only, as we are seeing the entire file object attached to the file field, also it is not going harm anything as this file object going to stored on `static` cache. :)

scottrigby:
Since this solves the issue in a way that doesn't hurt performance, I'm going to go ahead and merge this :shipit: 

If in the future someone can find the root cause of the file object not being fully loaded, and has an idea to address it in a better way, we can open another d.o issue on to do that 😄

PR #52, opened 2016-04-21
Issue 2711193: Move instant article object alter hook from render to a getter

scottrigby:
Resolves https://www.drupal.org/node/2711193

### DIFF
--- a/modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
+++ b/modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
@@ -63,12 +63,9 @@ function facebook_instant_articles_api_remove_action($node, $context = array()) 
  */
 function fb_instant_articles_api_rules_import_article($node) {
   if (isset($node->nid)) {
-    $layout_settings = fb_instant_articles_display_get_node_layout_settings($node->type);
-    $wrapper = \Drupal\fb_instant_articles_display\DrupalInstantArticleDisplay::create($node, $layout_settings);
-    $node->fb_instant_articles_display_wrapper = $wrapper;
     node_build_content($node, $view_mode = 'fb_instant_article');
     $client = \Drupal\fb_instant_articles_api\DrupalClient::get();
-    $client->importArticle($wrapper->instantArticle, $node->status === NODE_PUBLISHED);
+    $client->importArticle($node->fb_instant_articles_display_wrapper->getArticle(), $node->status === NODE_PUBLISHED);
   }
 }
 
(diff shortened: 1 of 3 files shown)


scottrigby:
Note that anbarasan.r reviewed this earlier but didn't have time to comment before running out (for commit credit if someone merges this before he comments).

scottrigby:
Also I discussed with @simonengelhardt

anbarasan1985:
This looks good to me and tested 👍

modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
```diff
@@ -63,12 +63,9 @@ function facebook_instant_articles_api_remove_action($node, $context = array())
  */
 function fb_instant_articles_api_rules_import_article($node) {
   if (isset($node->nid)) {
-    $layout_settings = fb_instant_articles_display_get_node_layout_settings($node->type);
-    $wrapper = \Drupal\fb_instant_articles_display\DrupalInstantArticleDisplay::create($node, $layout_settings);
```
simonengelhardt, line 67: Were these lines simply redundant because they also happen in [`fb_instant_articles_display_node_load`](https://github.com/BurdaMagazinOrg/module-fb_instant_articles/blob/7.x-1.x/modules/fb_instant_articles_display/fb_instant_articles_display.module#L188) (just trying to understand)?

anbarasan1985:
@simonengelhardt `fb_instant_articles_display_node_load` fires only when we explicitly call `node_load` for the node object but `fb_instant_articles_api_rules_import_article` can be called in some other places like other hooks.

simonengelhardt:
@anbarasan1985 @scottrigby In that case, do we still need to create the wrapper instance? Though I'm not sure if someone would be able to call with a `$node` without having gone through `node_load` (in which case we definitely don't need to instantiate the wrapper again)?

m4olivei:
Looks good to me 👍

scottrigby:
@simonengelhardt that's exactly my thought. As long as we ensure that hook_node_load() has fired then a bare display wrapper object will have had to already been created with the node layout settings, and stashed in the node. We were re-doing that work unnecessarily.

scottrigby:
@m4olivei cool thanks we had only tested with the API module, but I tried to be thorough in searching for existing calls go the public $instantArticle var, or the wrapper render(). Anyone who feels like merging & marking the d.o issue fixed, please feel free :)

simonengelhardt:
👍 :shipit:

PR #54, opened 2016-04-26
Issue #2713951: Add missing xmlns:namespace declaration to the views based RSS feeds.

m4olivei:
Resolves https://www.drupal.org/node/2713951.


simonengelhardt:
lgtm :shipit:

PR #55, opened 2016-04-27
Issue #2703063: Refactor base classes for extensibility

scottrigby:
Here is an alternate approach to #35, blending the singleton wrapper and Drupal hook patterns to solve extensibility limitations imposed by the SDK class design.

There are API changes: Drupal hook changes are documented in `fb_instant_article.api.php`. The rest are notes in the commit messages, and ultimately will be documented in [drupal.org/node/2703063](https://drupal.org/node/2703063) issue summary, as well as in the next release tag notes.


scottrigby:
Also note data model changes:
- `$entity->fb_instant_articles_display_wrapper` (contained the Display wrapper class), has now changed to `$entity->fb_instant_article_wrapper` (now containing the Drupal-hook-altered SDK `InstantArticle` object retrieved from `ArticleWrapper`).

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -5,6 +5,134 @@
...
+/**
+ * Implements hook_field_formatter_view().
+ */
+function fb_instant_articles_display_field_formatter_view($entity_type, $entity, $field, $instance, $langcode, $items, $display) {
+  // Grab the InstantArticle object we're going to be building up. If it's not
+  // present, it means that a site admin selected an FB IA field formatter for
+  // a non-FB IA content type.
+  if (!isset($entity->fb_instant_article_wrapper)) {
+    return FALSE;
+  }
```
m4olivei, line 48: Nice

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -5,6 +5,134 @@
...
+  foreach (array_keys(entity_get_info()) as $entity_type) {
+    $code = '';
+    if (!function_exists("fb_instant_articles_display_preprocess_{$entity_type}")) {
+      $code .= 'function fb_instant_articles_display_preprocess_' . $entity_type . "(&" . '$vars' . ") {
+      if (" . '$vars' . "['view_mode'] === 'fb_instant_article') {
+        " . '$vars' . "['theme_hook_suggestions'][] = 'entity__fb_instant_article';
+      }
+    }";
+    }
+    eval($code);
```
m4olivei, line 105: This is kinda neat.  However, couldn't we implement `hook_preprocess()`, and match the `$hook` param against entity_type machine names got from `entity_get_info()`?  That way we could avoid the `eval`.

scottrigby:
Status of general Entity support (not only node):
- The Base module doesn't know anything about Entity/Bundle mapping at all.
- The API module also doesn't care about Entity mapping.
- The Display module supports any Entity type, with a few UI exceptions:
  - The UI to select the bundle as an Instant Article type is only on the node type form (there is a `@todo` for splitting the guts of those functions to make it easier to add this UI support to other entity types).
  - The UI for the view mode layout only adds the node Title for node bundles (there is a `@todo` in `fb_instant_articles_display_get_fields()` for supporting other Entity types).
  - However, for both of these cases, in D7 there is not generic entity type form support, except in the contrib Entity API module. This may be worth exploring.
- The RSS module still only supports node (this can be a follow-up issue).
- The Views module still only supports node (this can be a follow-up issue).
- API Rules module still only supports node (this can be a follow-up issue). Note that I'm not aware of a solid, Entity-agnostic way to trigger `node_build_content()` (needed for triggering `hook_field_formatter_view()` properly), except for the contrib Entity API module's `entity_view()` function, which would make this task a lot easier and more stable.

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby, line 105: @m4olivei I would love that. But `hook_preprocess()` docs say:

> It is called for all theme hooks implemented as templates, **but not for theme hooks implemented as functions**. hook_preprocess_HOOK() can be used to preprocess variables for a specific theme hook, whether implemented as a template or function.

Honestly, I was wondering if we ought to skip this theme hijacking altogether, and add Views support for `Entity rendered as Instant Article`, rather than the default rendered entity option, and create a handler that just runs `$entity->fb_instant_article_wrapper->getArticle()->render('<!doctype html>', TRUE);` (it would let us remove a lot of code in the existing module).

scottrigby, line 105: That could also be a follow-up issue to explore that idea? But for now, this generic entity method has parity with the node-only functionality in the module now. What do you think?

m4olivei:
Just finished reading through.  First impressions are, wow, this is awesome!  I really like it.  When time allows I'll test this out properly.  A bit swamped ATM.  Thanks @scottrigby!

modules/fb_instant_articles_display/fb_instant_articles_display.module
m4olivei, line 105: Yeah I like that idea.  I've never done that with views, but it makes a lot of sense and would clean things up.

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -5,6 +5,134 @@
...
+  // Grab the InstantArticle object we're going to be building up. If it's not
+  // present, it means that a site admin selected an FB IA field formatter for
+  // a non-FB IA content type.
+  if (!isset($entity->fb_instant_article_wrapper)) {
+    return FALSE;
+  }
+
+  $article = $entity->fb_instant_article_wrapper->getArticle();
+  $entity_info = entity_get_info($entity_type);
+  $bundle_key = !empty($entity_info['entity keys']['bundle']) ? $entity_info['entity keys']['bundle'] : $entity_type;
```
m4olivei: Can we use `entity_extract_ids` instead here?

src/ArticleWrapper.php
```diff
@@ -0,0 +1,90 @@
...
+   *
+   * Also invokes a hook to allow other modules to alter the InstantArticle
+   * object before render or any other operation.
+   *
+   * @see hook_fb_instant_articles_instant_article_alter()
+   *
+   * @return \Facebook\InstantArticles\Elements\InstantArticle
+   */
+  public function getArticle() {
+    drupal_alter('fb_instant_articles_article', $this->instantArticle);
```
m4olivei: Hum, this changes the original intent of the hook a bit too much.  The former version passed along the relevant $node object, which could be used as context to alter the instant article.  Is there a way to pass along the `$node` object from here?  Or maybe some other way around this like what if `Drupal\fb_instant_articles_display\EntityWrapper` extended `Drupal\fb_instant_articles\ArticleWrapper` somehow and overrode this method to pass the additional context?

scottrigby: @m4olivei you're right, I actually meant to add a `$context` param to the hook as well. This will allow modules to add arbitrary contextual information to the `ArticleWrapper` for altering the InstantArticle object. Hang on and I'll add a commit for this.

scottrigby: @m4olivei Check out 24a1188 and f9f3ea9.

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby: @m4olivei Yep good call. I had already fixed this in another test I'm doing. Just added it to this PR in a897974.

scottrigby:
@m4olivei I made what I think is an improvement in b4d4771. The commit message and comments explain why I did this. Can you take a look again (i've addressed your concerns above), and especially at this last commit, because it works toward separating out the formatters ([field formatter are plugins in D8](https://www.drupal.org/node/1805846), so if we start updating the 8.x branch from this one, we will be closer to code parity with this change).

anbarasan1985:
@scottrigby Overall this looks good to me, did you get a chance to test this on our demo site?

scottrigby:
OK, I found a serious issue with `hook_fb_instant_articles_instant_article_alter()` being invoked inside `ArticleWrapper->getArticle()`: The Display module's `fb_instant_articles_display_field_formatter_view()` will call that for EVERY field FBIA formatter, which means the hook will get invoked over and over. This would mean developers implementing this hook would need to use `drupal_static()` with an array like `$articles[$entity_type][$entity]` if they want to make use of that without adding an element more than once for the same Entity (not only is that bad DX, it negates the reason for decoupling this base class from the Display module's 1:1 tie to an Entity). SO instead, I simplified this greatly in 98ee457, backing out of the `$context` class var, setter and getter, in favor of:

More API Changes:
- Moved 
  `hook_fb_instant_articles_instant_article_alter()` invocation into `ArticleWrapper::create()`.
- Introduced new `$context` param to `ArticleWrapper::create()`

This has solved the issues, and is ready for testing.

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -5,6 +5,132 @@
...
+    if (!in_array($type, array_keys($enabled_entity_types))
+      || !in_array($bundle, array_keys($enabled_entity_types[$type]))) {
+      return FALSE;
+    }
+
+    // We add Entity-specific information to the ArticleWrapper object, and
+    // attach it to the Entity for use in later hooks. Note that we don't store
+    // the InstantArticle object on the node directly, because our Singleton
+    // wrapper + Drupal hook class pattern relies on invoking
+    // hook_fb_instant_articles_article_alter() in ArticleWrapper->getArticle(),
```
m4olivei: Comment needs to be updated, since the alter happens in `ArticleWrapper::__construct` now.

src/ArticleWrapper.php
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * Instantiates the InstantArticle object, adding Drupal Base module configs
+   * where appropriate. Also invokes a hook to allow other modules to alter the
+   * InstantArticle object before render or any other operation.
+   *
+   * @param array $context
+   *   An associative array of contextual information altering the
+   *   InstantArticle object.
+   *
+   * @see hook_fb_instant_articles_instant_article_alter()
```
m4olivei: Should be `hook_ fb_instant_articles_article_alter`

modules/fb_instant_articles_display/src/EntityPropertyMapper.php
```diff
@@ -0,0 +1,209 @@
...
+        Time::create(Time::MODIFIED)
+          ->withDatetime(
+            \DateTime::createFromFormat('U', $this->entity->changed)
+          )
+      );
+    }
+
+    // Default the article author to the username.
+    if (isset($this->entity->uid)) {
+      $author = user_load($this->entity->uid);
```
m4olivei, line 114: With the change to running `hook_fb_instant_articles_article_alter` on create, I wonder if that will make the author field on the instant article impossible to override?  Since this will come after the alter has run and set it.  Will test.

m4olivei, line 114: Yeah, with the order of the hook now, this overwrites any change to the author that an alter hook might make. 🙍

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -5,6 +5,132 @@
...
+    // to allow other modules to alter the InstantArticle object whenever it is
+    // requested. If we request that now, it will not have the Entity-specific
+    // information on it this module needs for later altering.
+    $context = array(
+      'entity_type' => $type,
+      'entity' => $entity,
+    );
+    $wrapper = \Drupal\fb_instant_articles\ArticleWrapper::create($context);
```
m4olivei: With the hook now running within the call to `ArticleWrapper::create`, if you have a `hook_fb_instant_articles_article_alter` implementation that calls `entity_load`, you trigger an infinite call loop that ends in a Fatal Max Nesting error.  That's b/c this calls the `hook_fb_instant_articles_article_alter` implemenation, which triggers a `entity_load`, which calls the `hook_fb_instant_articles_article_alter`, which triggers an `entity_load`, etc.

I think with this issue and the other issue I noted regarding not being able to override the author, we need to find a place for the alter hook that comes after the Display module is done changing things on the `InstantArticle` object..

src/ArticleWrapper.php
scottrigby: @m4olivei Nice catch, fixed in 7eaf694.

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby: @m4olivei Excellend catch. This is why I make comments like this 😄  Now that we've moved the alter hook to `ArticleWrapper::__construct()` the reason I called `EntityPropertyMapper::create()` in `hook_entity_load()` is now moot, and even why we store `ArticleWrapper` instead of `InstantArticle` on the Entity is now not needed. In 089e621 I changed strategy to storing the `InstantArticle` object directly onto the Entity. I also now call `EntityPropertyMapper` from `hook_fb_instant_articles_article_alter()` instead of `hook_entity_load()` (and modified our `hook_module_implements_alter()` so our `hook_fb_instant_articles_article_alter()` run first, giving modules the ability to alter our default property mappings in `EntityPropertyMapper::map()`). 

Note that none of this allows us an alter hook to modify the `InstantArticles` object after all the _field_ mappings are finished (I can see no way to allow that, and also follow this architecture, so that we get the value from the Display module for API usage)… but I guess a custom module could implement their own `hook_field_formatter_view()`, ensure it runs after the Display module implementation, and change the default mappings from `EntityFieldMapper::map()` that way.

modules/fb_instant_articles_display/src/EntityPropertyMapper.php
scottrigby, line 114: @m4olivei Good point. I've pushed a fix for this (see https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/55#discussion_r62128704). The example I added in `hook_fb_instant_articles_article_alter()` works now.

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -5,6 +5,151 @@
...
+  }
+}
+
+/**
+ * Implements hook_fb_instant_articles_article_alter().
+ *
+ * @see fb_instant_articles_display_module_implements_alter()
+ */
+function fb_instant_articles_display_fb_instant_articles_article_alter($instantArticle, $context) {
+  if (!isset($context['entity_type']) || isset($context['entity'])) {
```
m4olivei: Second `isset` is missing the boolean not operator.

m4olivei:
The problem of the author not being able to be overwritten is fixed, however, the issue remains where if you have a call to `entity_load` in your implementation of `hook_fb_instant_articles_article_alter` you can trigger a call loop that results in a Fatal Maximum function nesting level reached error.  We for example `node_load` related articles based on a field value to tack that on to InstantArticle Footer.  I'm not sure how to get around this, but it's feeling more and more like kicking off the InstantArticle stuff in the `entity_load` is the wrong place.  The other thing I realized is that the creation of the instant article happens in `hook_entity_load` regardless of view mode, so for example if you have FBIA enabled for the blog content type, that code will run any time that node is rendered, even on the front end website as a teaser or something.  Which is unnecessary work.  That issue is already present in the existing codebase too I realize.  Your suggestion about implementing Views support for Entity rendered as Instant Article is feeling relevant here too.

Thoughts?

scottrigby:
@m4olivei OK yeah, let's separate these issues:

### 1. Recursive loop when calling `entity_load()` within `hook_fb_instant_articles_article_alter()`:

This can happen with various hooks - for example, if you try to call `entity_view()` within `hook_field_formatter_view()` you'll similarly get an infinite loop (because `field_attach_view()` re-calls the latter, whenever you call the former. A more obvious example is calling `node_load()` inside `hook_node_load()`). But if developers have a good reason to do this, they can just do a custom Entity type and ID check, like this:

``` php
function hook_fb_instant_articles_article_alter($instantArticle, $context) {
  $target_type = 'user';
  $target_id = 9999;
  if ($target_type != $context['entity_type'] && $target_id != $id) {
     $entities = entity_load($target_type, array($target_id));
    // Now do whatever you want with your Entity without worrying about recursion.
  }
}
```

I just sketched out an example we could add to our API docs to explain how to do this. Check out the diff just below. If you think this is not unnecessarily complex, we could include it in this PR (or maybe add to a 'Troubleshooting' section of the Display module's README?).

``` diff
diff --git a/fb_instant_article.api.php b/fb_instant_article.api.php
index 30600f4..eae3aa7 100644
--- a/fb_instant_article.api.php
+++ b/fb_instant_article.api.php
@@ -31,7 +31,7 @@ function hook_fb_instant_articles_article_alter($instantArticle, $context) {
   // Check to see if a module like fb_instant_articles_display has set Entity
   // context.
   if (isset($context['entity_type']) && isset($context['entity'])) {
-    // Example: Set the author using a custom callback.
+    // Example 1: Set the author using a simple custom callback.
     $byline = custom_fb_instant_articles_byline_callback($context['entity_type'], $context['entity']);
     $instantArticle
       ->getHeader()
@@ -39,6 +39,40 @@ function hook_fb_instant_articles_article_alter($instantArticle, $context) {
         \Facebook\InstantArticles\Elements\Author::create()
           ->withName($byline)
       );
+
+    // Example 2: Add additional authors from a value on a referenced Entity.
+    list($id) = entity_extract_ids($context['entity_type'], $context['entity']);
+    // In this example, we use the label of each target Entity from an Entity
+    // Reference field.
+    $field_name = 'field_author_reference';
+    $field = field_info_field($field_name);
+    $target_type = $field['settings']['target_type'];
+    if ($items = field_get_items($context['entity_type'], $context['entity'], $field_name)) {
+      foreach($items as $item) {
+        $target_id = $item['target_id'];
+        // Note that it is good practice to check that the Entity type and ID of
+        // the Entity you plan to load is not the same as the current Entity. If
+        // you have the fb_instant_articles_display module enabled, this check
+        // is required in order to prevent entity_load() recursion. The reason
+        // is that it's implementation of hook_entity_load() creates an
+        // InstantArticle object, which invokes this hook.
+        if ($target_type != $context['entity_type'] && $target_id != $id) {
+          $target_ids[] = $target_id;
+        }
+      }
+      if (!empty($target_ids)) {
+        $entities = entity_load($target_type, $target_ids);
+        foreach ($entities as $entity) {
+          $byline = entity_label($target_type, $entity);
+        }
+        $instantArticle
+          ->getHeader()
+          ->addAuthor(
+            \Facebook\InstantArticles\Elements\Author::create()
+              ->withName($byline)
+          );
+      }
+    }
   }
 }
```

### 2. InstantArticle creation happens in `hook_entity_load()` regardless of view mode (problem exists in current module too).

I think this is by design:  Our idea was that other modules making use of the Display module can have flexibility about which Entities are recognized as an InstantArticle, without needing any custom code (for instance, the Views module can use any filters it wishes, and the API Rules module can use any conditions it wishes) - and of course, sites that define this in custom code can use their own logic (but that should not be necessary).

And the addition of a bare InstantArticle object is pretty light as we have it… But I do get what you mean: right now, a module that wants to lighten this has no way to do so. If we want to allow this, we could let the Display module limit individual Entities that are _not_ InstantArticles (even within the check we already do for the configured Entity type and Bundle that we already do). We could invoke a new hook like `hook_fb_instant_articles_display_repress_article_load($entity_type, $entity)` that allows other modules to define whether or not the Entity is an instant article before creating a new InstantArticle object and attaching it to the Entity. I'm thinking this may not really be worth the trouble though, because one way or another we're invoking a hook anyway, but I guess we could benchmark how much not creating a bare the InstantArticle object could save? 

Either way, I think this should be a follow-up issue for discussion, since this already happens in the current code, right?

modules/fb_instant_articles_display/fb_instant_articles_display.module
scottrigby: @m4olivei Thanks, fixed in 6998ce5.

scottrigby:
@m4olivei OK, you can skip my last comment. Tl;dr, I decided it was cleaner to add in a "Troubleshooting" section of a README for the Display module, because it's only applicable if this optional module is enabled, and that hook documentation is in the Base module (and people should RTFM if it's available). Also this module didn't have a README (we have an issue for that at http://drupal.org/node/2704731 so I numbered that accordingly in the last commit b19659e). Since this resolves the first half of your last comment above and is trivial to implement, and since the second half of your last comment is already a problem in the module (so it should be a follow up), do you have any further concerns with merging this PR? I'd like to get this refactoring settled before making a `7.x-1.0` release, and also before updating the `8.x-1.x` branch to match `7.x-1.x` (ideally during DrupalCon this week).

scottrigby:
I also just added the Transformer hook from #35 in f1526bf (see this comment thread https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/35/files#r59240081). That was originally in the scope of this issue, but I just remembered I didn't include that in this PR.

modules/fb_instant_articles_display/fb_instant_articles_display.module
```diff
@@ -5,6 +5,151 @@
...
+ * @see fb_instant_articles_display_module_implements_alter()
+ * @see fb_instant_articles_display_theme()
+ * @see theme_entity__fb_instant_article()
+ */
+function fb_instant_articles_display_declare_entity_preprocess_hooks() {
+  foreach (array_keys(entity_get_info()) as $entity_type) {
+    $code = '';
+    if (!function_exists("fb_instant_articles_display_preprocess_{$entity_type}")) {
+      $code .= 'function fb_instant_articles_display_preprocess_' . $entity_type . "(&" . '$vars' . ") {
+      if (" . '$vars' . "['view_mode'] === 'fb_instant_article') {
```
m4olivei: For some entities `$vars['view_mode']` is undefined so we get PHP notices.  Needs an `isset` call.

src/TransformerExtender.php
```diff
@@ -0,0 +1,66 @@
...
+   * new invokeRules() method - because of the way Transformer is normally used.
+   *
+   * @see hook_fb_instant_articles_transformer_rules()
+   *
+   * Note the parent class does not have a constructor, so we do not call
+   * @code parent::__construct() @endcode.
+   */
+  public function __construct() {
+    $rules = module_invoke_all('fb_instant_articles_transformer_rules');
+    $this->addRules($rules);
```
m4olivei, line 35: As discussed with @scottrigby, let's add an drupal_alter to allow modules to alter the final list of rules across all implementations of the info hook.

scottrigby, line 35: Added in e6cb795.

m4olivei:
This is looking good.  As discussed though, we've now got a couple breaking changes such that it's worth bumping the major version of the module to 7.x-2.x.  @scottrigby to open a new 7.x-2.x branch, close this PR and reopen against that new branch.

scottrigby:
Awesome @m4olivei - I'm closing this in favor of #60 so we can build the `7.x-2.x` branch.

PR #56, opened 2016-04-27
Issue #2714683 : Getting Fatal error: Call to a member function getArticle()

anbarasan1985:
Moved the DrupalInstantArticleDisplay class built to `hook_node_view` instead of `hook_node_load`. Because `hook_node_load` won't be fired when we adding the content. Like node insert. Since we are using the `node_build_content` or `node_view` in API module and RSS module it makes me feel appropriate to move this. 
This is reference to #52

### DIFF
--- a/modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
+++ b/modules/fb_instant_articles_api_rules/fb_instant_articles_api_rules.module
@@ -63,6 +63,9 @@ function facebook_instant_articles_api_remove_action($node, $context = array()) 
  */
 function fb_instant_articles_api_rules_import_article($node) {
   if (isset($node->nid)) {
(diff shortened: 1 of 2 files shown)


m4olivei:
Discussed with @scottrigby, but to re-iterate here, there is a problem using `hook_node_view` b/c of the order that hooks are called. I believe it's b/c `hook_node_view` only gets called after all of the field formatters are called, at which point it’s too late to init the shared object.  I just tried it with the views submodule and what you end up with is empty article contents.  Like any of the fields configured on the manage display settings page don't render.

anbarasan1985:
@m4olivei TY for testing this with views. I updated `fb_instant_articles_api_rules.module` to fix this issue.

scottrigby:
Code review 👍 , and tested. Let's :shipit:

PR #59, opened 2016-05-09
Issue #2703377: Simplify hook_html_head_alter()

scottrigby:
We don't need to call `drupal_add_html_head()` when using `hook_html_head_alter()`. Also simplify the nesting by re-using the same `if` condition for the FB page ID variable check.


m4olivei:
👍

PR #63, opened 2016-05-16
Issue #2726761: add a default rule to transform div's to paragraphs to cover a common case and avoid an empty article transform.

m4olivei:
Resolves https://www.drupal.org/node/2726761.


scottrigby:
Same reasoning as #64 (although there is not a good way to alter default rules in `7.x-1.x`, this should be the default behavior. If devs want to alter it, they should be using `7.x-2.x` anyway).

PR #64, opened 2016-05-16
Issue #2726761: add a default rule to transform div's to paragraphs to cover a common case and avoid an empty article transform.  Adapted for 7.x-2.x.

m4olivei:
Resolves https://www.drupal.org/node/2726761 for 7.x-2.x


scottrigby:
I agree this should be the default transformation for `<div>`. 👍  Note that in `7.x-2.x`, there is an alter rule if anyone doesn't want this for some custom reason.

PR #66, opened 2016-05-26
Issue #2734403: rss feed fix node operation

logmiller:
https://www.drupal.org/node/2734403


logmiller:
Push an commit that would also fix rss feeds checkbox in the forms from being always set to be enabled (https://www.drupal.org/node/2735299).

scottrigby:
@logmiller looks like you're mixing issues. Let's keep each d.o matched to a separate PR. e78907d is ready to merge whenever you rebase out the other commit. I'm not so sure about issue 2735299 yet, let's discuss separately.

PR #68, opened 2016-05-31
Issue #2734403: RSS Feeds: invalid callback in node operations

logmiller:
Addressed concern with previous pull request.
Isolated both issues into separate branches.

Resolves issue: https://www.drupal.org/node/2734403


scottrigby:
👍

PR #69, opened 2016-05-31
Issue #2735299: RSS Feeds: toggling the checkbox to "Include in fia rss feed"

logmiller:
Addressed concern with previous pull request.
Isolated both issues into separate branches.

Resolves issue: https://www.drupal.org/node/2735299


modules/fb_instant_articles_rss/fb_instant_articles_rss.module
```diff
@@ -178,49 +178,42 @@ function fb_instant_articles_rss_form_node_form_alter(&$form, &$form_state, $for
...
 
     // Is an Article?
-    $type = $form['#node']->type;
-    if (!empty($form['#node']->nid)) {
-      $fb_instant_enabled = fb_instant_articles_rss_is_article($form['#node']->nid);
-    }
-    else {
-      $fb_instant_enabled = fb_instant_articles_display_is_article_type('node', $type);
-    }
+    $is_type = fb_instant_articles_display_is_article_type('node', $form['#node']->type);
```
m4olivei, line 181: This check is already made in the outer `if` statement, so you can assume it's `TRUE`, and don't need to do it again.

modules/fb_instant_articles_rss/fb_instant_articles_rss.module
```diff
@@ -178,49 +178,42 @@ function fb_instant_articles_rss_form_node_form_alter(&$form, &$form_state, $for
...
-      '#description' => t('Enable this entity to be included in the Facebook Instant Articles RSS feed.'),
-      '#weight' => 0,
-      '#default_value' => ($previously_checked || $fb_instant_enabled) ? TRUE : FALSE,
-    );
+      $form['fb_instant_articles_rss']['fb_instant_enabled'] = array(
+        '#type' => 'checkbox',
+        '#title' => t('Include in Facebook Instant Articles RSS feed.'),
+        '#description' => t('Enable this entity to be included in the Facebook Instant Articles RSS feed.'),
+        '#weight' => 0,
+        '#default_value' => ($fb_instant_enabled) ? TRUE : FALSE,
```
m4olivei, line 194: Simplify to `'#default_value' => $fb_instant_enabled`

modules/fb_instant_articles_rss/fb_instant_articles_rss.module
```diff
@@ -178,49 +178,42 @@ function fb_instant_articles_rss_form_node_form_alter(&$form, &$form_state, $for
...
+  if (!isset($form_state['values']['fb_instant_enabled'])) {
+    $form_state['values']['fb_instant_enabled'] = 0;
+  }
+  $value = $form_state['values']['fb_instant_enabled'];
+  if ($value) {
+    // Save the new article entity setting
+    fb_instant_articles_rss_set_entity($form['nid']['#value'], $value);
+  }
+  else {
+    // delete the entry to keep the table clean.
```
m4olivei, line 216: Capitalize "delete".

m4olivei: Wouldn't we need to store a value of 0 to explicitly say that this should be disabled? In case the default value changed from 0 to 1?

m4olivei (changes requested):
Thanks for the PR, left a few comments.

m4olivei:
Closing as there hasn't been any activity in awhile.

PR #72, opened 2016-06-16
Issue #2749119 by ryanissamson: Adds variable that overrides base canonical url

ryanissamson:
- If a url is supplied, the base portion is overridden.
- This could probably be improved with validation on
  the field to prevent malformed urls.

https://www.drupal.org/node/2749119


scottrigby:
This looks correct 👍  Most comments so far are in the d.o issue.

PR #76, opened 2016-06-16
Issue #2750353: Add check to fieldFormatVideoElement in case $video_url is provided to function

ryanissamson:
```
Adds a check for URL instead of just URI

- We were altering a reference field and the data passed in
already had a URL and no URI, so this just prevents this
failing the check.
```


m4olivei:
Closing as this hasn't seen any activity in awhile.

PR #78, opened 2016-08-12
Fixed strpos() expects parameter 1 to be string.

markdorison:
Handle array or string values in #theme element.

@see https://www.drupal.org/node/2780753


scottrigby:
@markdorison 👍  :shipit:

PR #79, opened 2016-08-18
Add possibility to add an Ad to the article header

tjwelde:


dbosen:
This looks fine to me. @scottrigby: would you be OK with merging this?

PR #80, opened 2016-09-05
Correct timestamp format:

NoDiskInDriveA:
- escape literal T delimiter as per TWIG documentation (double backslashes)
- use timezone identifier instead of literal Z


NoDiskInDriveA:
The timestamps in the feed currently evaluate to something like "2016-05-09CEST14:53:327200" - this happens b/c the literals in the format string are not correctly escaped. This commit fixes this.

I also changed the literal "Z" (Zulu) at the end of the timestamp to the timezone identifier of the current date context. While I could not find any details about the required date format in the FB IA documentation, we cannot simply assume the site's timezone is always GMT. Using the actual timezone instead of a hardcoded Z seems to be quickest fix here.

PR #82, opened 2016-09-07
Fix warnings thrown in Drupal 8.1.8

tjwelde:
These warnings are thrown, because the contentnamespace can not be found.


m4olivei:
The 8.x-1.x branch is no longer supported. Please try using the 8.x-2.x branch.

PR #90, opened 2017-05-02
Issue #2871632 port the base settings config form to Drupal 8

m4olivei:
Ports the base module settings for to Drupal 8.

Also a bunch of code style updates in here, namely changing to PHP short array syntax.

Things to check:
 - [x] All settings ported over.
 - [x] Test the admin form, change all settings. Try inputting invalid data for the ads fields to ensure validation works.
 - [x] Save a FB page_id and check the source on the home page to ensure the metatag renders.


fb_instant_articles.module
```diff
@@ -8,6 +8,38 @@
 use Drupal\Core\Routing\RouteMatchInterface;
 
 /**
+ * No ads will automatically be inserted into Instant Articles.
```
deviantintegral: Can we put these constants in a class somewhere? Same thing with the method below.

fb_instant_articles.module
```diff
@@ -46,24 +116,24 @@ function fb_instant_articles_page_attachments_alter(array &$page) {
  * Implements hook_init().
  */
 function fb_instant_articles_init() {
-  $enable_transformer_logging = \Drupal::config('fb_instant_articles_enable_logging');
-  $appender = array(
+  $enable_transformer_logging = \Drupal::config('fb_instant_articles.base_settings')->get('enable_logging');
+  $appender = [
     'class' => $enable_transformer_logging ? '\Drupal\fb_instant_articles\DrupalLoggerAppender' : 'LoggerAppenderNull',
```
deviantintegral, line 126: Double check, is `LogAppenderNull` supposed to be a full class name, or is it just a special keyword?

src/Form/BaseSettingsForm.php
```diff
@@ -0,0 +1,263 @@
...
+    // Add the page id configuration.
+    $args = [
+      '@claim_url' => 'https://developers.facebook.com/docs/instant-articles/claim-url',
+    ];
+    $form['page_id'] = [
+      '#title' => t('Facebook Page ID'),
+      '#type' => 'textfield',
+      '#default_value' => $config->get('page_id'),
+      '#required' => TRUE,
+      '#description' => $this->t('In order to designate the domain that will host your
```
deviantintegral: I'm pretty sure this has to all be on one line for localize.drupal.org to be able to extract the string.

src/Form/BaseSettingsForm.php
```diff
@@ -0,0 +1,263 @@
...
+    if ($form_state->getValue('ads_type') != FB_INSTANT_ARTICLES_AD_TYPE_SOURCE_URL) {
+      return;
+    }
+
+    $ads_iframe_url = $form_state->getValue('ads_iframe_url');
+    if (empty($ads_iframe_url)) {
+      $form_state->setErrorByName('ads_iframe_url', $this->t('You must specify a valid source URL for your Ads when using the Source URL ad type.'));
+    }
+
+    if (filter_var($ads_iframe_url, FILTER_VALIDATE_URL) === FALSE) {
```
deviantintegral: We should probably use `\Drupal\Component\Utility\UrlHelper::isValid` here.

src/Form/BaseSettingsForm.php
```diff
@@ -0,0 +1,263 @@
...
+      return;
+    }
+
+    $ads_an_placement_id = $form_state->getValue('ads_an_placement_id');
+    if (empty($ads_an_placement_id)) {
+      $form_state->setErrorByName('ads_an_placement_id', $this->t('You must specify a valid Placement ID when using the Audience Network ad type.'));
+    }
+
+    if (preg_match('/^[\d_]+$/', $ads_an_placement_id) !== 1) {
+      $form_state->setErrorByName('ads_an_placement_id', $this->t('You must specify a valid Placement ID when using the Audience Network ad type.'));
```
deviantintegral: Does it help to call out what the placement ID is or where you can find it?

deviantintegral (changes requested):
This is well on the way! Needs tests and some relatively minor changes.

bgbohling:
D'oh!  Forgot to do the push after committing test code.  Tests are there now, but can probably use some tweaking.  They concentrate on verifying minimal functionality instead of complete coverage.

fb_instant_articles.module
bgbohling: @deviantintegral Is there a convention to follow for this, like class name, type, etc.?

fb_instant_articles.module
bgbohling: LoggerAppenders are a FB SDK thing that provides a way to forward log messages from the SDK to elsewheres, like Drupal watchdog.  So I'm going to guess that `LoggerAppenderNull` is a keyword.

src/Form/BaseSettingsForm.php
bgbohling: Made it a single line.

src/Form/BaseSettingsForm.php
bgbohling: OK, using that thing, then.

src/Form/BaseSettingsForm.php
bgbohling: It looks like this is something set up in the brand's Audience Network account for Instant Articles.  Adding the instrux from the FB developer docs to the form feedback.

fb_instant_articles.module
m4olivei: @deviantintegral we can, but at this stage of the port there isn't a good home for it yet. Let's leave it for now and come back to it. These constants will most likely see actual use in https://www.drupal.org/node/2872807. There should be a good home in the implementation there. I'd like to wait until that evolves, in the meantime, I'll mark a @ todo.

fb_instant_articles.module
m4olivei: It is a class name. It lives in the apache/log4php package. It's in the global namespace, and this value is used within the package, so I believe it's correct as is.

m4olivei:
Addressed all review comments. Added some extra assertions, rebased against 8.x-2.x, and cleaned up some code style.

@deviantintegral could you give this another look?

bgbohling (approved):
Looks good to me.

fb_instant_articles.module
deviantintegral: Boo, global namespace. Should it be prefixed with a \ then to make it clear, given that the other option is namespaced?

deviantintegral (approved):

deviantintegral:
The backslash thing is no big deal, whatever you decide this is good to merge. Thanks!

m4olivei:
Thanks!

PR #91, opened 2017-05-04
Issue #2872801 Add a custom view mode for Facebook Instant Articles

m4olivei:
Adds the view mode for Facebook Instant Articles. In so doing starts the port for `fb_instant_articles_display` module. Also introduces a toggle per bundle as a helper to enable the fb_instant_articles view mode.

To test:

 - [ ] Enable `fb_instant_articles_display` module.
 - [ ] Verify you get the Facebook Instant Articles view mode on node types
 - [ ] On the node bundle edit page, click the Facebook Instant Articles vertical tab, and enable the checkbox. This should enable/disable the Facebook Instant Articles view mode for the bundle.

PR #93, opened 2017-05-18
Issue #2872801 Add custom view mode to the base module.

m4olivei:
Adds the view mode for Facebook Instant Articles. Introduces a toggle per bundle as a helper to enable the fb_instant_articles view mode.

To test:

 - [ ] Enable `fb_instant_articles` module. If already enabled, you'll need to uninstall, reinstall.
 - [ ] Verify you get the Facebook Instant Articles view mode on node types
 - [ ] On the node bundle edit page, click the Facebook Instant Articles vertical tab, and enable the checkbox. This should enable/disable the Facebook Instant Articles view mode for the bundle.

### DIFF
--- a/config/install/core.entity_view_mode.node.facebook_instant_articles.yml
+++ b/config/install/core.entity_view_mode.node.facebook_instant_articles.yml
@@ -1,0 +1,12 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+  enforced:
+    module:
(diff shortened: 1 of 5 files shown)


m4olivei:
Alternate to #91. Adds view mode support to the core module rather than a fb_instant_articles_display module.

bgbohling (approved):
Reviewed and tested.   Works as expected.

PR #94, opened 2017-05-19
Issue #2872804 Customize the Manage Display UI for the custom view mode to allow multiple regions

m4olivei:
This PR adds the customizations to Manage Display UI for the FB Instant Articles view mode. To test:

 - [x] Enable the fb_instant_articles
 - [x] Toggle FB Instant Articles on for the Article content type for example
 - [x] Visit the Manage Display UI for the Article content type. You should see Header/Body/Footer regions. You should be able to drag fields into each, Save and the fields should persist in those regions.

### DIFF
--- a/src/Form/EntityViewDisplayEditForm.php
+++ b/src/Form/EntityViewDisplayEditForm.php
@@ -3,11 +3,25 @@
 namespace Drupal\fb_instant_articles\Form;
 
 use Drupal\field_ui\Form\EntityViewDisplayEditForm as CoreEntityViewDisplayEditForm;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Field\FieldDefinitionInterface;
 
 /**
  * Extends the core EntityViewDisplayEditForm to support multiple regions.
  */
 class EntityViewDisplayEditForm extends CoreEntityViewDisplayEditForm {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getTableHeader() {
+    $table_headers = parent::getTableHeader();
+    if ($this->getEntity()->getOriginalMode() === 'fb_instant_articles') {
+      $region = $this->t('Region');
+      array_splice($table_headers, 1, 0, array($region));
+    }
+    return $table_headers;
+  }
 
   /**
    * {@inheritdoc}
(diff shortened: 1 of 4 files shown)


src/Form/EntityViewDisplayEditForm.php
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Extends the core EntityViewDisplayEditForm to support multiple regions.
+ */
+class EntityViewDisplayEditForm extends CoreEntityViewDisplayEditForm {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getTableHeader() {
```
m4olivei: @bgbohling I'm not following why this was added. Was something not working as expected?

src/Form/EntityViewDisplayEditForm.php
```diff
@@ -0,0 +1,85 @@
...
+      $new_regions['hidden'] = $regions['hidden'];
+      $regions = $new_regions;
+    }
+    return $regions;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function buildFieldRow(FieldDefinitionInterface $field_definition, array $form, FormStateInterface $form_state) {
```
m4olivei: @bgbohling same here. I'm not following why this was added. Was something not working as expected?

src/Form/EntityViewDisplayEditForm.php
bgbohling: I added this and buildFieldRow() because the FBIA display was only showing the same columns and rows as all the other display types.  This was what it took for me to get the Regions column to show up in the FBIA display settings.

m4olivei: I see. The thing is that the regions column is implicit in where the field is dragged to, it sits under a clearly labeled region name. If a user wanted to use the drop downs to manage it, they can click the "Show Row Weights" link at the top right of the table, and the hidden columns all show up. I think we should leave it like that, it's a common pattern throughout these forms. OK if I revert it?

bgbohling: Sure.  All I know is that the form didn't have all the things it needed until I added these bits.

m4olivei: True, I realize that from the screencaps I put in the ticket, however, I think leaving them out is actually better for D8, since it natively supports the concept we are after.

bgbohling (approved):

PR #96, opened 2017-06-09
Issue #2872808: Implement field formatter for each FBIA element

m4olivei:
This adds field formatters for all the FBIA elements we had in the Drupal 7 module. The list includes:

* Ads
* Analytics
* Author
* Blockquote
* Copyright
* Credits
* Image
* Interactive
* Kicker
* List
* Pullquote
* Social
* Subtitle
* Transformer
* Video

This PR simply makes the field formatters available for use. It doesn't do anything to start building the Instant Article, that will come in a later issue. In terms of review, please do a static code review as usual and then check the following:

* You can assign "FBIA" prefixed field formatters to fields in the Mange Display UI for the Facebook Instant Articles display mode, eg. `/admin/structure/types/manage/article/display/fb_instant_articles`
* Assign each formatter in the list above and if applicable try their settings to ensure they work.
* The "FBIA" prefixed field formatters do not show up as an option in other display modes, eg. `/admin/structure/types/manage/article/display/teaser`

### DIFF
--- a/src/Plugin/Field/FieldFormatter/FormatterBase.php
+++ b/src/Plugin/Field/FieldFormatter/FormatterBase.php
@@ -3,12 +3,12 @@
 namespace Drupal\fb_instant_articles\Plugin\Field\FieldFormatter;
 
 use Drupal\Core\Field\FieldItemListInterface;
-use Drupal\Core\Field\FormatterBase;
+use Drupal\Core\Field\FormatterBase as DrupalFormatterBase;
 
 /**
  * Base class for all of our FBIA field formatters.
  */
-class FbiaFormatterBase extends FormatterBase {
+class FormatterBase extends DrupalFormatterBase {
 
   const SOURCE_TYPE_URL = 'url';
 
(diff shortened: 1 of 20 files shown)


src/Plugin/Field/FieldFormatter/FormatterBase.php
```diff
@@ -0,0 +1,27 @@
...
+
+  const SOURCE_TYPE_HTML = 'html';
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
+    // Do nothing. Our field formatters are unique in that they are not meant
+    // to generate HTML on their own, but only signal to the Serialization API
+    // the fate of this field in the FBIA document.
```
bgbohling, line 23: How does the Serialization API use these?  Is that happening in DRACO-2522?

scottrigby (approved):
Sick

src/Plugin/Field/FieldFormatter/FormatterBase.php
m4olivei: Good question! The way it will work is the Serialization API is given the entity (usually a node) as an argument. It then iterates over the configured fields, each field being an object of type `FieldItemListInterface`. It then calls the field formatter with that object as the first param, so that's how it'll get it. This method signature will change a bit too in https://www.drupal.org/node/2872807.

PR #97, opened 2017-06-14
Issue #2871642 Implement light object wrappers around FB SDK Client object to use Drupal configured credentials

m4olivei:
Adds a service that wraps the FB SDK client to use the configured Drupal credentials.

This isn't yet used anywhere, but you can verify it's working by first connecting with Facebook via the API configuration form, then opening up `drush php` or equivalent and trying:

```php
$client = \Drupal::service('fb_instant_articles.drupal_client');
$client->getArticleIDFromCanonicalURL('http://example.com')
```

You should get a `null` response, and not an exception.


bgbohling (approved):
Like that factory creation trick.

PR #99, opened 2017-06-20
Issue #2871639 Implement views Row plugin for rendering each node entity as an RSS feed item

m4olivei:
This (finally) gets us something to submit to Facebook. This PR puts together the RSS feed spec'd by Facebook ready to submit to the Facebook Instant Articles platform.

To test manually:

 - [ ] Install the `fb_instant_articles_views` module (if you already have it installed uninstall it first).
 - [ ] Verify you get the view that the module ships with (`/admin/structure/views/view/facebook_instant_articles_rss`)
 - [ ] Test the view and ensure that you get valid RSS feed output per the FBIA spec: https://developers.facebook.com/docs/instant-articles/publishing/setup-rss-feed

It looks like we'll need this on a public server before we can test this in Facebook, so for now, verifying the format of the RSS feed will suffice.


bgbohling (approved):

m4olivei:
Made some changes to handle debug output better (it was getting in the way of a valid feed when the Transformer was configured for use). Also included `<channel>` child elements that I was missing.

m4olivei:
@bgbohling still OK, any questions/concerns?

PR #100, opened 2017-06-20
2871640 views filter plugin

bgbohling:


modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
```diff
@@ -14,15 +9,15 @@
...
+ * Simple filter that checks if a node implements the FIA custom view mode.
  *
  * @ingroup views_filter_handlers
  *
  * @ViewsFilter("validfacebookinstantarticles")
  */
 class ValidFacebookInstantArticles extends FilterPluginBase {
 
-  const FIA_VIEW_MODE = 'fb_instant_articles_views_rss';
+  const FIA_VIEW_MODE = 'fb_instant_articles';
```
m4olivei: I'd like to find a better place for this constant, since it's used elsewhere. Perhaps `\Drupal\fb_instant_articles\Form\EntityViewDisplayForm`.

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
```diff
@@ -34,13 +29,20 @@ public function adminSummary() {
...
+  public function canBuildGroup() {
+    return FALSE;
+  }
+
   /**
    * Provide the basic form which calls through to subforms.
+   *
    * If overridden, it is best to call through to the parent,
    * or to at least make sure all of the functions in this form
    * are called.
```
m4olivei: This method description doesn't make a lot of sense, or at least isn't adding any value. I get the sense it's been copied from somewhere. I think it's better to just change this to `{@inheritdoc}`.

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
```diff
@@ -64,68 +66,70 @@ public function init(ViewExecutable $view, DisplayPluginBase $display, array &$o
...
 
   /**
-   * Set the values for this filter to all node bundles that implement custom settings for the view mode
+   * Set the values for this filter.
    *
-   * This uses an unreliable method for detecting custom settings for node bundles.  The method is to check for a config
-   * Entity for the entity display settings, for the bundle/view mode.  Essentially this looks for a yml file/record
-   * for the custom settings, and one is found, the bundle is considered FIA active.
+   * This applies to all node bundles that implement custom settings for the
+   * view mode.
```
m4olivei: Change to "for the **fb_instant_articles** view mode"

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
```diff
@@ -64,68 +66,70 @@ public function init(ViewExecutable $view, DisplayPluginBase $display, array &$o
...
    *
+   * This uses an unreliable method for detecting custom settings for node
```
m4olivei: As far as I know what's done here is in fact a reliable method of finding this out. I think we can nix this paragraph.

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
```diff
@@ -64,68 +66,70 @@ public function init(ViewExecutable $view, DisplayPluginBase $display, array &$o
...
    */
   protected function enabledNodeBundlesSetValues() {
-    /**
-     * @var \Drupal\Core\Entity\EntityTypeBundleInfo $entityBundleInfo
+    /*
+     * @var \Drupal\Core\Entity\EntityTypeBundleInfo $entity_bundle_info
      *  entity_type.bundle.info
      */
-    $entityBundleInfo = \Drupal::service('entity_type.bundle.info');
+    $entity_bundle_info = \Drupal::service('entity_type.bundle.info');
```
m4olivei: Use dependency injection to get this service instead.

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
```diff
@@ -64,68 +66,70 @@ public function init(ViewExecutable $view, DisplayPluginBase $display, array &$o
...
 
-    /**
-     * @var \Drupal\Core\Entity\EntityStorageInterface $entityStorage
+    /*
+     * @var \Drupal\Core\Entity\EntityStorageInterface $entity_storage
      *   View Mode entity storage handler
      */
-    $entityStorage = \Drupal::service('entity_type.manager')->getStorage('entity_view_display');
+    $entity_storage = \Drupal::service('entity_type.manager')->getStorage('entity_view_display');
```
m4olivei: Same as above, use dependency injection to get this service instead.

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
```diff
@@ -64,68 +66,70 @@ public function init(ViewExecutable $view, DisplayPluginBase $display, array &$o
...
+      /*
+       * @var \Drupal\Core\Entity\EntityInterface|null $view_mode
        *   Config entity for the view mode, if it exists
        */
-      $viewMode = $entityStorage->load($viewModeId);
+      $view_mode = $entity_storage->load($view_mode_id);
 
-      if ($viewMode instanceof EntityInterface) {
-        $nodeTypes[$id] = $id;
+      if ($view_mode instanceof EntityInterface) {
```
m4olivei: Might make more sense here to check for an instance of `EntityViewDisplayInterface`, because that's what we're really interested in checking here.

m4olivei (changes requested):

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
bgbohling: ✔️

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
bgbohling: Nixed.

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
bgbohling: Sounds logical, but I'm not able to make that work.  Ended up with `node_field_data.type = ''` in the query.

m4olivei: That could be expected behaviour if you don't have any content types enabled for FBIA.

bgbohling: But I do, and with EntityInterface, the type shows up in the query.

m4olivei: Really? It works for me. I just stepped through in xdebug and `$view_mode` ends up being a `\Drupal\Core\Entity\Entity\EntityViewDisplay` object which is an instance of `\Drupal\Core\Entity\Display\EntityViewDisplayInterface`...

bgbohling: D'oh!  Gotta remember to `use` the interface.

modules/fb_instant_articles_views/src/Plugin/views/filter/ValidFacebookInstantArticles.php
bgbohling: Drat.  The create()/__construct() thing isn't working here, so tried a factory like you did for DrupalClient.  What a surprise, can't get that working, either.

m4olivei (approved):

m4olivei:
@bgbohling if this all looks good to you, we can merge.

PR #102, opened 2017-06-23
Issue #2871646 hook_entity_save support for import and delete API operations

m4olivei:
Integrates with the Instant Articles API to import and delete articles using entity hooks.

This also includes some fixes to the RSS approach to ensure the canonical URL override is used correctly.

To test:

 - [ ] Enable `fb_instant_articles_api` module.
 - [ ] Enable a content type for FBIA
 - [ ] Make sure your API settings are configured.
 - [ ] Create/edit a FBIA node, save it, it should be sent to Facebook

For more detailed testing instructions, see this Paper document: https://paper.dropbox.com/doc/Testing-Drupal-FBIA-Integration-FgIQeFRcEur3FdzOtDvV2

### DIFF
--- a/modules/fb_instant_articles_api/fb_instant_articles_api.info.yml
+++ b/modules/fb_instant_articles_api/fb_instant_articles_api.info.yml
@@ -1,6 +1,6 @@
 name: Facebook Instant Articles API
 type: module
-description: Implements entity hooks that use the Instant Articles API to delver articles to the Instant Article platform.
+description: Implements entity hooks that use the Instant Articles API to deliver articles to the Instant Article platform.
 core: 8.x
 package: Facebook Instant Articles
 dependencies:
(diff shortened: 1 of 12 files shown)


m4olivei:
Found a bug with this where if you try and Save a node enabled for Facebook Instant Articles before the API settings have been configured, it blows up. Should skip attempting to send to Facebook if the API credentials have not been configured yet.

m4olivei:
Fixed.

modules/fb_instant_articles_api/fb_instant_articles_api.info.yml
```diff
@@ -1,6 +1,6 @@
 name: Facebook Instant Articles API
 type: module
-description: API module for Facebook Instant Articles.
+description: Implements entity hooks that use the Instant Articles API to delver articles to the Instant Article platform.
```
bgbohling: Spellcheck nazi:  delver should be deliver (since this is going out to the Drupal world at large).

bgbohling:
Just not seeing anything else that looks wrong to me.

modules/fb_instant_articles_api/fb_instant_articles_api.info.yml
m4olivei: Nice laser eyes. Fixed in most recent commit.

PR #105, opened 2017-07-04
Issue #2889771 Enabling IA on a content type at the same time as creating the content type throws and error

m4olivei:
Fixes the error that was thrown when a new content type was saved with the FBIA checkbox checked.

To test:

 - Create a new content type and set the "Enable Facebook Instant Articles" checkbox.
 - Save the content type
 - If you didn't get an error, yay! It works.
 - Run the test suite to verify there are no regressions.
 
Note, there is also a functional test included in this to verify the fix as well.

### DIFF
--- a/fb_instant_articles.module
+++ b/fb_instant_articles.module
@@ -109,7 +109,13 @@ function fb_instant_articles_form_node_type_form_alter(&$form, FormStateInterfac
     '#default_value' => $node_type->getThirdPartySetting('fb_instant_articles', 'fb_instant_articles_enabled'),
   ];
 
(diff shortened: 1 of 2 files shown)


bgbohling:
Drat. 

bbohling$ git checkout 2889771-missing-bundle-entity
error: pathspec '2889771-missing-bundle-entity' did not match any file(s) known to git.

???

m4olivei:
@bgbohling try `git fetch` first.

bgbohling (approved):
Works as expected.  Don't see anything scary in code.

m4olivei:
@bgbohling awesome! Could you mark the Drupal issue as RTBC?

PR #106, opened 2017-07-07
Issue #2892840 Removing a field that was configured on the fb_instant_articles view mode throws an error when serializing a node

m4olivei:
Check the steps to reproduce from the issue and make sure you don't get any errors.


bgbohling (approved):
Good catch.

PR #107, opened 2017-07-07
Issue #2892435 FBIA transformer eliminates non-h1,h2 headers

m4olivei:
To test:

* Checkout this branch
* Clear Drupal cache
* Check the RSS feed for a node with h3-h6 elements. They should all be present as h2 elements in the Instant Article markup.


bgbohling (approved):
Interesting that they don't allow those, but hey.

PR #108, opened 2017-07-13
Issue #2893626 Image and video does not render correctly when using non-FBIA formatters

m4olivei:
To test:

* Checkout branch
* Configure an image/file field on a FBIA enable content type and place it in Header/Body of the FBIA view mode.
* Create a node of this type.
* View the RSS feed and ensure that image is in the feed.


bgbohling (approved):
Don't see anything wrong with the single line change, and it's working on my local (images showing up in the RSS feed).

PR #109, opened 2017-07-13
Issue #2893620 Use Canonical URL override field when set for image/video/file field types

m4olivei:
To Test:

* Checkout branch
* Set a value for "Canonical URL override" in the Base Settings form (/admin/config/services/fb_instant_articles).
* Configure an image/file field on a FBIA enable content type and place it in Header/Body of the FBIA view mode.
* Create a node of this type.
* View Source on the RSS feed and ensure that the image you added is using the Canonical URL override.

### DIFF
--- a/src/Plugin/Field/FieldFormatter/ImageFormatter.php
+++ b/src/Plugin/Field/FieldFormatter/ImageFormatter.php
@@ -2,8 +2,12 @@
 
 namespace Drupal\fb_instant_articles\Plugin\Field\FieldFormatter;
 
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\Core\Field\FieldItemListInterface;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Session\AccountInterface;
 use Drupal\fb_instant_articles\Plugin\Field\InstantArticleFormatterInterface;
 use Drupal\fb_instant_articles\Regions;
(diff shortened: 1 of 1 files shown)


src/Plugin/Field/FieldFormatter/ImageFormatter.php
```diff
@@ -26,6 +31,39 @@
...
+   */
+  public function __construct($plugin_id, $plugin_definition, FieldDefinitionInterface $field_definition, array $settings, $label, $view_mode, array $third_party_settings, AccountInterface $current_user, EntityStorageInterface $image_style_storage, ConfigFactoryInterface $config_factory) {
+    parent::__construct($plugin_id, $plugin_definition, $field_definition, $settings, $label, $view_mode, $third_party_settings, $current_user, $image_style_storage);
+    $this->config = $config_factory->get('fb_instant_articles.settings');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
```
bgbohling, line 51: Just for curiosity, are $configuration, $plugin and $plugin_definition here FBIA things?

bgbohling (approved):
This is working as expected on my local.

src/Plugin/Field/FieldFormatter/ImageFormatter.php
m4olivei: Nope, nothing specific to FBIA. This class defines a Field Formatter plugin, and field formatter plugins support dependency injection via this create method. Here I'm overriding the parent implementation, which you can find in `\Drupal\image\Plugin\Field\FieldFormatter\ImageFormatter`. Most of the stuff here is satisfying that implementation so as not to break things. This is a common pattern you'll find throughout Drupal in plugins.

PR #110, opened 2017-07-14
Issue #2894871 Entity reference Label field formatter does not display in the output

m4olivei:
See issue for the problem, should no longer be reproducible.


scottrigby (approved):
LGTM!

bgbohling:
Working on my local.

PR #112, opened 2017-07-21
Issue #2896337 Nodes are not excluded by the views filter even though Facebook Instant Articles is turned off

m4olivei:
*To test*

* Checkout branch
* Toggle support for Facebook Instant Articles on on the Content type edit page (eg. `/admin/structure/types/manage/article`), and Save it, such that the FBIA view mode is enabled.
* Create some content of that type and verify it shows in the RSS feed (eg. `/instant-articles.rss`)
* Toggle FBIA off on the same content type.
* Visit `/instant-article.rss`, you should see no content of that type in the views output.


bgbohling:
Working as expected.

bgbohling (approved):

PR #113, opened 2017-07-24
Issue #2896986 FBIA video is not showing up in RSS feed

m4olivei:
Fix an issue where Canonical URL was not taken into account with the FBIA video formatter.

**To test**

* Setup a file field on an FBIA enabled content type.
* Configure the FBIA view mode and set that file field to use FBIA video.
* Create content with a value for that file field
* Set a canonical URL in the settings (eg. `/admin/config/services/fb_instant_articles`)
* View the RSS feed (eg. `/instant-articles.rss`) and view source. You should see the URL to the file you added and it should be using the canonical URL you set in settings.

### DIFF
--- a/tests/src/Kernel/Plugin/Field/FieldFormatter/VideoFormatterTest.php
+++ b/tests/src/Kernel/Plugin/Field/FieldFormatter/VideoFormatterTest.php
@@ -123,14 +123,14 @@ class VideoFormatterTest extends FormatterTestBase {
 
     // Test with a canonical URL set.
     $config = $this->config('fb_instant_articles.settings');
-    $config->set('canonical_url_override', 'http://example.com')
+    $config->set('canonical_url_override', 'http://example.com/override')
       ->save();
     /** @var \Drupal\fb_instant_articles\Plugin\Field\InstantArticleFormatterInterface $formatter */
     $formatter = $this->display->getRenderer($this->fieldName);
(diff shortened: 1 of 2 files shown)


tests/src/Kernel/Plugin/Field/FieldFormatter/VideoFormatterTest.php
```diff
@@ -106,4 +110,27 @@ public function testVideoFormatter() {
...
+   */
+  function testVideoFormatterCanonicalUrl() {
+    $entity = EntityTest::create([]);
+    // Handy method to populate the field with a real value.
+    // @see FileItem::generateSampleValue()
+    $entity->{$this->fieldName}->generateSampleItems(1);
+
+    // Test with a canonical URL set.
+    $config = $this->config('fb_instant_articles.settings');
+    $config->set('canonical_url_override', 'http://example.com')
```
deviantintegral: `example.com` is pretty common, and I could see this test passing by mistake. How about making the override `http://example.com/override`?

m4olivei: Sure thing! Fixed.

deviantintegral (approved):
Only done a static review, but looks great!

bgbohling:
Working on my local, too.

bgbohling (approved):

PR #120, opened 2017-10-11
Issue #2914843: Fix PHP error for unallowed array values in class constants

hctom:
... in FieldItemListNormalizer class

See issue on drupal.org for details: https://www.drupal.org/node/2914843


m4olivei:
Looks great thanks!

PR #121, opened 2017-10-12
Issue #2915661: Make field item list normalization fallback region-aware

hctom:
... so transformation happens with the correct region context element

See Drupal.org issue: https://www.drupal.org/node/2915661

### DIFF
--- a/src/Normalizer/FieldItemListNormalizer.php
+++ b/src/Normalizer/FieldItemListNormalizer.php
@@ -95,14 +95,14 @@ class FieldItemListNormalizer extends SerializerAwareNormalizer implements Norma
           @$document->loadHTML(Html::decodeEntities($markup));
 
           // Determine correct context for transformation.
-          $context = $article;
+          $transformer_context = $article;
           if ($component['region'] === Regions::REGION_HEADER) {
             $header = $article->getHeader();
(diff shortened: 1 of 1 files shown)


src/Normalizer/FieldItemListNormalizer.php
```diff
@@ -91,8 +94,27 @@ public function normalize($object, $format = NULL, array $context = []) {
           $markup = '<!doctype html><html><head><meta charset="utf-8" /></head><body>' . $markup . '</body></html>';
           @$document->loadHTML(Html::decodeEntities($markup));
 
-          // Note that we are putting the content of these fields into the body.
-          $this->transformer->transform($article, $document);
+          // Determine correct context for transformation.
+          $context = $article;
```
m4olivei: Let's call it something other than `$context`, since that is already used as the third argument to this method. Maybe `$article_context`?

m4olivei (changes requested):
This is a great addition. Just one small comment. Also, it would be great to have test coverage for this, want to take a shot at adding tests?

hctom:
You are definitely right, I just commited a change, where the new `$context`variable is renamed to `$transformer_context`

m4olivei (approved):

PR #122, opened 2017-10-12
Issue #2915243 RSS feed doesn't invalidate when changing display

bgbohling:
Override save() method in EntityViewDisplayEditForm to invalidate FBIA rss cache tag when form is saved

### DIFF
--- a/src/Form/EntityViewDisplayEditForm.php
+++ b/src/Form/EntityViewDisplayEditForm.php
@@ -5,6 +5,8 @@ namespace Drupal\fb_instant_articles\Form;
 use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\fb_instant_articles\Regions;
 use Drupal\field_ui\Form\EntityViewDisplayEditForm as CoreEntityViewDisplayEditForm;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Cache\Cache;
 
 /**
  * Extends the core EntityViewDisplayEditForm to support multiple regions.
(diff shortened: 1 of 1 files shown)


src/Form/EntityViewDisplayEditForm.php
```diff
@@ -57,4 +59,13 @@ protected function getApplicablePluginOptions(FieldDefinitionInterface $field_de
     return $options;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function save(array $form, FormStateInterface $form_state) {
+    // Way to get the fbia rss cache tag dynamically from here?
+    Cache::invalidateTags(['config:views.view.facebook_instant_articles_rss']);
```
m4olivei, line 67: This is a good idea, however, there are a couple thing that precludes us from using an approach like this. First this class overrides _all_ Manage display forms (not just for the FBIA view mode), so we're effectively invalidating the FBIA RSS view anytime any Mange display form is saved. Second, it's entirely possible that a site builder would decide to create an additional view that outputs a FBIA RSS feed, which would be named something other than the one we ship with the module (`facebook_instant_articles_rss`). If that were the case, those additional views would exhibit the same issue we're trying to fix.

I think we need to instead take the reverse approach, which is when the RSS feed is being generated, we should declare a dependency (cache tag) on the FBIA view mode for the relevant bundle that is being output. That was when the view mode settings are changed, Drupal will automatically invalidate the views output. The other thing to keep in mind is that the module also allows you the flexibility to mix node bundles (Articles and Blog posts) in the same feed. So we can only know which entity view mode displays we are dependent on as we are rendering the RSS feed.

I've got something I think will work. I'm going to submit an alternate PR.

m4olivei (changes requested):
There are a couple issues with this approach. See my inline comment. I've put together an alternate solution which should be more robust and cover all our bases.

m4olivei:
Closed in favour of https://github.com/BurdaMagazinOrg/module-fb_instant_articles/pull/123.

PR #123, opened 2017-10-13
Issue #2915243 Declare the correct cache tag dependency during the RSS views output.

m4olivei:
To test:

* Make sure Drupal's Page Cache and Dynamic Page Cache are on. Also helps to have the cache headers enabled.
* Enable the Facebook Instant Articles Views module and configure Article content type. Set the image to use FBIA Image, and enable Likes and Comments.
* Create and article, making sure to set an Image.
* View the feed (`/instant-articles.rss`) and note that likes and comments are enabled for the image of each article by viewing the source.
* Refresh a couple times and, if you have cache headers enabled, note that `X-Drupal-Cache-Tags` should contain a value of `config:core.entity_view_display.node.article.fb_instant_articles`. It should also show as cached with a header of `X-Drupal-Dynamic-Cache: HIT` and/or `X-Drupal-Cache: HIT`.
* Change the FBIA view mode to uncheck Likes and Comments on the FBIA Image formatter.
* Refresh the feed and you should see the updates right away, without having to manually clear the cache.

### DIFF
--- a/src/Normalizer/InstantArticleContentEntityNormalizer.php
+++ b/src/Normalizer/InstantArticleContentEntityNormalizer.php
@@ -144,6 +144,11 @@ class InstantArticleContentEntityNormalizer extends SerializerAwareNormalizer im
     // If we're given an entity_view_display object as context, use that as a
     // mapping to guide the normalization.
     if ($display = $this->entityViewDisplay($data, $context)) {
+      // Declare a dependency on the view mode configuration if we are rendering
+      // in the context of a views REST export.
+      if (isset($context['views_style_plugin'])) {
+        $context['views_style_plugin']->displayHandler->display['cache_metadata']['tags'] = array_merge($context['views_style_plugin']->displayHandler->display['cache_metadata']['tags'], $display->getCacheTags());
+      }
       $context['entity_view_display'] = $display;
       $components = $this->getApplicableComponents($display);
       uasort($components, [$this, 'sortComponents']);
--- a/src/Normalizer/InstantArticleContentEntityNormalizer.php
+++ b/src/Normalizer/InstantArticleContentEntityNormalizer.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\fb_instant_articles\Normalizer;
 
+use Drupal\Core\Cache\Cache;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Entity\ContentEntityInterface;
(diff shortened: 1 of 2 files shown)


bgbohling (approved):

src/Normalizer/InstantArticleContentEntityNormalizer.php
```diff
@@ -144,6 +144,11 @@ public function normalize($data, $format = NULL, array $context = []) {
     // If we're given an entity_view_display object as context, use that as a
     // mapping to guide the normalization.
     if ($display = $this->entityViewDisplay($data, $context)) {
+      // Declare a dependency on the view mode configuration if we are rendering
+      // in the context of a views REST export.
+      if (isset($context['views_style_plugin'])) {
+        $context['views_style_plugin']->displayHandler->display['cache_metadata']['tags'] = array_merge($context['views_style_plugin']->displayHandler->display['cache_metadata']['tags'], $display->getCacheTags());
```
davereid: I think you should possibly use [Cache::mergeTags()](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Cache%21Cache.php/function/Cache%3A%3AmergeTags/8.2.x) instead of array_merge()? I would normally recommend using [CacheableMetadata::merge()](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Cache%21CacheableMetadata.php/function/CacheableMetadata%3A%3Amerge/8.2.x) but I think the Views style plugin's display handler probably doesn't implement `\Drupal\Core\Cache\CacheableMetadata`.

src/Normalizer/InstantArticleContentEntityNormalizer.php
```diff
@@ -144,6 +145,11 @@ public function normalize($data, $format = NULL, array $context = []) {
     // If we're given an entity_view_display object as context, use that as a
     // mapping to guide the normalization.
     if ($display = $this->entityViewDisplay($data, $context)) {
+      // Declare a dependency on the view mode configuration if we are rendering
+      // in the context of a views REST export.
+      if (isset($context['views_style_plugin'])) {
+        $context['views_style_plugin']->displayHandler->display['cache_metadata']['tags'] = Cache::mergeTags($context['views_style_plugin']->displayHandler->display['cache_metadata']['tags'], $display->getCacheTags());
```
bgbohling, line 151: Yep, I'd never have thought of that in a million years. ;)

PR #130, opened 2018-03-01
Issue #2941425 Pin FBIA PHP SDK to 1.7.0

m4olivei:
1.7.0 was the last version before the logger was refactored and `symfony/css-selector` was narrowly required at `2.8.*`. Since Drupal > 8.4 requires `symfony/css-selector` at `~3.2.8` composer needs to go back to version 1.7.0 anyways. This also resolves an issue with the Logger being refactored in version `1.7.1`, which allows the module to be used in Drupal 8.3.x, which was more liberal about `symfony/css-selector`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "homepage": "https://www.drupal.org/node/2676800/committers"
   }],
   "require": {
-    "facebook/facebook-instant-articles-sdk-php": "^1.0.0"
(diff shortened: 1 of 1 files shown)


deviantintegral:
Is there an upstream issue for these b/c breaks? Or was it a bug before in that the upstream constraint was too loose?

m4olivei:
Indeed! I filed this upstream issue: https://github.com/facebook/facebook-instant-articles-sdk-php/issues/321

PR #135, opened 2018-08-23
Issue #2994506 Class hierarchy ties fbia_rss normalizer to fbia normalizer such that it can't be overriden

m4olivei:
Resolves https://www.drupal.org/project/fb_instant_articles/issues/2994506

### DIFF
--- a/src/Normalizer/EntityHelperTrait.php
+++ b/src/Normalizer/EntityHelperTrait.php
@@ -19,6 +19,19 @@ trait EntityHelperTrait {
   protected $config;
 
   /**
+   * Convenience method for getting the right config object.
+   *
+   * @return \Drupal\Core\Config\ImmutableConfig
+   *   Facebook Instant Articles config object.
+   */
+  protected function config() {
+    if (!$this->config) {
+      $this->config = \Drupal::config('fb_instant_articles.settings');
+    }
+    return $this->config;
+  }
+
+  /**
    * Helper function to compute the canonical URL for a given entity.
    *
    * @param \Drupal\Core\Entity\ContentEntityInterface $entity
@@ -30,7 +43,7 @@ trait EntityHelperTrait {
    * @throws \Drupal\Core\Entity\EntityMalformedException
    */
   public function entityCanonicalUrl(ContentEntityInterface $entity) {
-    if ($override = $this->config->get('canonical_url_override')) {
+    if ($override = $this->config()->get('canonical_url_override')) {
       return $override . $entity->toUrl('canonical')->toString();
     }
     else {
--- a/src/Normalizer/EntityHelperTrait.php
+++ b/src/Normalizer/EntityHelperTrait.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace Drupal\fb_instant_articles\Normalizer;
+
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityChangedInterface;
+use Drupal\user\EntityOwnerInterface;
+
+/**
+ * Common entity data getters shared between the normalizers.
+ */
+trait EntityHelperTrait {
+
+  /**
+   * Instant articles config.
+   *
+   * @var \Drupal\Core\Config\ImmutableConfig
+   */
+  protected $config;
+
+  /**
+   * Helper function to compute the canonical URL for a given entity.
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   Content entity being normalized.
+   *
+   * @return string
+   *   The canonical URL for the given entity.
+   *
+   * @throws \Drupal\Core\Entity\EntityMalformedException
+   */
+  public function entityCanonicalUrl(ContentEntityInterface $entity) {
+    if ($override = $this->config->get('canonical_url_override')) {
+      return $override . $entity->toUrl('canonical')->toString();
+    }
+    else {
+      return $entity->toUrl('canonical', ['absolute' => TRUE])->toString();
+    }
+  }
+
+  /**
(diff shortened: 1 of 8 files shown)


src/Normalizer/InstantArticleContentEntityNormalizer.php
```diff
@@ -202,31 +197,15 @@ protected function entityViewDisplay(ContentEntityInterface $entity, array $cont
...
-   * Helper function to compute the canonical URL for a given entity.
-   *
-   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
-   *   Content entity being normalized.
-   *
-   * @return string
-   *   The canonical URL for the given entity.
-   */
-  public function entityCanonicalUrl(ContentEntityInterface $entity) {
-    if ($override = $this->config->get('canonical_url_override')) {
```
vijaycs85: we could add linkTemplate to entities via [hook_entity_type_alter](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21entity.api.php/function/hook_entity_type_alter/8.5.x) instead of keeping at config. Something like [this](http://cgit.drupalcode.org/drupal/tree/core/modules/settings_tray/settings_tray.module?h=8.5.x#n99)

src/Normalizer/EntityHelperTrait.php
```diff
@@ -0,0 +1,87 @@
...
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   Content entity being normalized.
+   *
+   * @return string
+   *   The canonical URL for the given entity.
+   *
+   * @throws \Drupal\Core\Entity\EntityMalformedException
+   */
+  public function entityCanonicalUrl(ContentEntityInterface $entity) {
+    if ($override = $this->config->get('canonical_url_override')) {
```
vijaycs85: we could add linkTemplate to entities via [hook_entity_type_alter](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21entity.api.php/function/hook_entity_type_alter/8.5.x) instead of keeping at config. Something like [this](http://cgit.drupalcode.org/drupal/tree/core/modules/settings_tray/settings_tray.module?h=8.5.x#n99)

src/Normalizer/FieldItemListNormalizer.php
```diff
@@ -53,14 +58,9 @@ public function __construct(RendererInterface $renderer, Transformer $transforme
...
-    $supported_formats = [
-      InstantArticleContentEntityNormalizer::FORMAT,
-      InstantArticleRssContentEntityNormalizer::FORMAT,
-    ];
-
     // Only consider this normalizer if we are trying to normalize a field item
-    // list into the 'fbia' or 'fbia_rss' format.
-    return in_array($format, $supported_formats) && $data instanceof FieldItemListInterface;
+    // list into the 'fbia' format.
+    return $format === self::FORMAT && $data instanceof FieldItemListInterface;
```
vijaycs85, line 63: I think it's a nice cleanup to use one format for both fields.

src/Normalizer/EntityHelperTrait.php
```diff
@@ -0,0 +1,87 @@
...
+ * Common entity data getters shared between the normalizers.
+ */
+trait EntityHelperTrait {
+
+  /**
+   * Instant articles config.
+   *
+   * @var \Drupal\Core\Config\ImmutableConfig
+   */
+  protected $config;
```
vijaycs85, line 19: Since it's a trait, probably we want to have `config()` so that if the parent class doesn't have it set, we can fallback?
```php
protected function config() {
  if (!$this->config) {
    $this->config = \Drupal::config('fb_instant_articles.settings');
  }
return $this->config;
}
```

src/Normalizer/InstantArticleRssContentEntityNormalizer.php
```diff
@@ -26,7 +39,7 @@ public function normalize($data, $format = NULL, array $context = []) {
       'title' => $data->label(),
       'link' => $this->entityCanonicalUrl($data),
       'guid' => $data->uuid(),
-      'content:encoded' => parent::normalize($data, $format, $context),
+      'content:encoded' => $this->serializer->normalize($data, 'fbia', $context),
```
vijaycs85, line 42: Same format to RSS as well!

vijaycs85 (approved):
Overall, it looks like a good improvement. Added few comments and questions.

src/Normalizer/EntityHelperTrait.php
m4olivei: Ooo, I like that. Will do.

src/Normalizer/InstantArticleRssContentEntityNormalizer.php
m4olivei: This is a bit of a special case here. The `fbia` normalizer "normalizes" to a InstantArticle object, and the the `fbia_rss` encoder looks for that object and runs the objects `render` method to dump the FBIA markup. I see your point, this could probably be done better, and I'm breaking the interface, but I'd like to defer any further refactoring down this road to a later point.

src/Normalizer/EntityHelperTrait.php
m4olivei: Oh that's interesting, TIL. Let me file a follow up issue for that. I think that could be worthwhile, but don't want to get too distracted at the moment.

m4olivei: Follow up issue: https://www.drupal.org/project/fb_instant_articles/issues/2994704

m4olivei:
Blarg, I forgot to credit you @vijaycs85. Sorry :(. Next time. Thanks for the review!

PR #139, opened 2019-04-01
Replace constructor with ::create()

alexpott:


src/Plugin/views/row/FiaFields.php
```diff
@@ -27,19 +26,14 @@ class FiaFields extends EntityRow {
...
-   * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
-   *   The language manager.
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, EntityManagerInterface $entity_manager, LanguageManagerInterface $language_manager) {
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
     $configuration['entity_type'] = 'node';
-    parent::__construct($configuration, $plugin_id, $plugin_definition, $entity_manager, $language_manager);
-    $this->options['view_mode'] = 'fb_instant_articles_rss';
+    $plugin = parent::create($container, $configuration, $plugin_id, $plugin_definition);
+    $plugin->options['view_mode'] = 'fb_instant_articles_rss';
```
chrfritsch, line 33: Shouldn't we set that like `EntityRow::defineOptions` does?

alexpott: Why not... I guess I just went for the closed thing to before

alexpott: Well here's one reason 
```
  protected function defineOptions() {
    $options = parent::defineOptions();
    $options['view_mode'] = ['default' => 'default'];
    return $options;
  }
```
defines view mode as an array but we're defining is as a string. And \Drupal\views\Plugin\views\row\EntityRow::summaryTitle() uses the value as an array key !?!?!?

I think we should keep close to the known good of HEAD and not mess too much here.

chrfritsch: Here is the explanaition why that works:

```  
protected function setOptionDefaults(array &$storage, array $options) {
    foreach ($options as $option => $definition) {
      if (isset($definition['contains'])) {
        $storage[$option] = [];
        $this->setOptionDefaults($storage[$option], $definition['contains']);
      }
      else {
        $storage[$option] = $definition['default'];
      }
    }
  }
```

But yeah, let's keep it close to the thing we had before.

PR #140, opened 2019-10-04
Issue #3085768: Enable serialization module on update

ol0lll:


m4olivei:
This is good stuff. Sorry it's taken so long. I'm going to move the review back to d.o.

PR #141, opened 2019-10-14
Issue #3087768: Missing config schema for rendered entity field formatter

ol0lll:


m4olivei:
This is good stuff. Sorry it's taken so long. I'm going to move the review back to d.o.

